PR #18, opened 2023-03-16
release: Goreleaser configuration and automation.

winder:
## Summary

Enable goreleaser for building and publishing conduit binaries.

The automation triggers whenever a tag starting with `v1.` is pushed.

The previous versioning and deployment stuff seemed really complicated, and I'm really not sure what it was adding. Please let me know if this smaller process is missing any important functionality that needs to be added back.

## Test Plan

I pushed a tag to my fork:
* [The action ran.](https://github.com/winder/conduit/actions/runs/4432232042)
* [A release was published. ](https://github.com/winder/conduit/releases/tag/v1.0.0)

The downloaded binary looks good:
```
$ ./conduit -v
Conduit 1.0.0 (c8262a5)
```

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -1,0 +1,30 @@
+name: goreleaser
+
+on:
+  push:
+    tags:
+      - 'v1.**'
+
+permissions:
+  contents: write
+  # packages: write
+  # issues: write
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - run: git fetch --force --tags
+      - uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
+      - uses: goreleaser/goreleaser-action@v4
+        with:
+          distribution: goreleaser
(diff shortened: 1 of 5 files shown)


Eric-Warehime (approved):
LGTM, but I'm not familiar with how the versioning bits work in Indexer so someone else should chime in on missing functionality.

algobarb:
I guess functionally, when comparing this release (created directly from github actions) and the indexer releases, we'll lose the **signed** binaries and the deb packages. However, you do include the checksums, so it depends on if this is enough.

This is not my strongest area, so I am just quoting from [this post](https://security.stackexchange.com/questions/102966/what-is-the-difference-between-the-sig-files-and-the-checksum-files-such-as-o):

> Checksums ensure [data integrity](https://en.wikipedia.org/wiki/Data_integrity).
> Digital signatures additionally ensure [data authenticity](https://en.wikipedia.org/wiki/Authentication).

One other thing we do that hasn't been automated is updating the release page as well: http://releases.algorand.com/

excalq (approved):
This seems like a useful utility, so I'm looking forward to trying in in our release process. I'm sure we'll want to make later additions. It should be useful bit of tooling towards reproducible builds.

PR #32, opened 2023-03-20
algod importer: Add catchup functionality to algod importer

Eric-Warehime:
## Summary

Adds fast catchup and slow catchup to the algod importer. Includes config options to specify a catchpoint and admin API key for algod to use for fast catchup. Also automatically tries to wait for the node to complete slow catchup to the proper sync round as specified by the pipeline init round.

## Test Plan

Adding tests incrementally in this PR.


conduit/plugins/tools/testutil/mockinit.go
```diff
@@ -13,6 +13,11 @@ func (m *MockInitProvider) GetGenesis() *sdk.Genesis {
 	return m.Genesis
 }
 
+// SetGenesis updates the genesis block in the init provider
+func (m *MockInitProvider) SetGenesis(genesis *sdk.Genesis) {
```
winder, line 17: IIRC This implementation is identical to the regular InitProvider but there is a different constructor function. Then there is a separate interface and implementation for the regular InitProvider. There was some complication with circular dependencies when I tried to consolidate them all.

Eric-Warehime: Removed it since they're the same. Left the interface in place so we can implement a new type if we need to differ the behavior ever.

conduit/init_provider.go
```diff
@@ -18,6 +18,11 @@ func MakePipelineInitProvider(currentRound *sdk.Round, genesis *sdk.Genesis) *Pi
 	}
 }
 
+// SetGenesis updates the genesis block in the init provider
+func (a *PipelineInitProvider) SetGenesis(genesis *sdk.Genesis) {
```
winder, line 22: This could probably be private so that nobody is tempted to call it outside of the pipeline package.

conduit/plugins/importers/algod/algod_importer.go
```diff
@@ -80,7 +81,88 @@ func init() {
...
+}
+
+func (algodImp *algodImporter) catchupNode(initProvider data.InitProvider) error {
+	// Set the sync round to the round provided by initProvider
+	_, err := algodImp.aclient.SetSyncRound(uint64(initProvider.NextDBRound())).Do(algodImp.ctx)
+	if err != nil {
+		return fmt.Errorf("received unexpected error setting sync round: %w", err)
+	}
+	// Run Catchpoint Catchup
+	if algodImp.cfg.CatchupConfig.Catchpoint != "" {
```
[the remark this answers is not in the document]
Eric-Warehime: I believe running fast catchup w/ a catchpoint ahead of the current DB round in the ledger will no-op regardless.

The follower node will also validate the catchpoint agains the sync round https://github.com/algorand/go-algorand/blob/master/node/follower_node.go#L349

The logic to parse the label doesn't exist outside of go-algorand either so we'd have to copy that into the SDK which seems like something we don't necessarily want copied into there.

Let me know your thoughts on this.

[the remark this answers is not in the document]
winder: I just gave this a try. If the node is on round 25M and you start a catchpoint at round 24M, the ledger will go backwards. So I'm not sure there's a way around having to check the round.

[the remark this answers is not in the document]
Eric-Warehime: You're right. I've added in the parsing of the round out of the catchpoint, and compared it against the node's latest round to make sure we're not spending a lot of time doing fast catchup just to go backwards.

The label round being ahead of where it should be is handled when we set `syncRound` to the next db round from the init provider before calling catchpoint catchup. So a catchup to a round after the next db round would be rejected based on the sync round.

Still adding unit testing for all of this.

conduit/init_provider.go
Eric-Warehime: This can't be private--it's used across a bunch of places.

conduit/plugins/importers/algod/algod_importer.go
```diff
@@ -134,6 +249,8 @@ func (algodImp *algodImporter) Init(ctx context.Context, cfg plugins.PluginConfi
 		return nil, fmt.Errorf("unable to fetch genesis file from API at %s", algodImp.cfg.NetAddr)
 	}
 
+	err = algodImp.catchupNode(initProvider)
```
shiqizng, line 252: when follower node's round equals  initProvider.NextDBRound(), no catchup is needed, right?

conduit/plugins/importers/algod/algod_importer.go
```diff
@@ -80,7 +82,120 @@ func init() {
...
+
+	}
+
+	algodImp.logger.Infof("Catchpoint catchup finished in %s", time.Since(start))
+	return nil
+}
+
+func (algodImp *algodImporter) catchupNode(initProvider data.InitProvider) error {
+	// Set the sync round to the round provided by initProvider
+	_, err := algodImp.aclient.SetSyncRound(uint64(initProvider.NextDBRound())).Do(algodImp.ctx)
```
shiqizng, line 159: this is also called for slow catchup?  when a follower node starts,  the sync round is set to latest+maxLookAhead. for a caught up node, this would move the sync round backward to latest+1?

conduit/pipeline/pipeline.go
```diff
@@ -264,14 +264,21 @@ func (p *pipelineImpl) Init() error {
...
+	// Initial genesis object is nil--gets updated after importer.Init
+	var initProvider data.InitProvider = conduit.MakePipelineInitProvider(&round, nil)
+	p.initProvider = &initProvider
+
 	configs, err := yaml.Marshal(p.cfg.Importer.Config)
 	if err != nil {
 		return fmt.Errorf("Pipeline.Start(): could not serialize Importer.Args: %w", err)
 	}
-	genesis, err := (*p.importer).Init(p.ctx, p.makeConfig("importer", importerName, configs), importerLogger)
+	genesis, err := (*p.importer).Init(p.ctx, *p.initProvider, p.makeConfig("importer", importerName, configs), importerLogger)
```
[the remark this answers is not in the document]
Eric-Warehime: I wanted to use the initProvider since it's also passed into the processors and exporters during Init. This would give us a uniform interface, although you're right all we need is the round so it would be simpler to pass that in instead.

conduit/plugins/importers/algod/algod_importer.go
Eric-Warehime: Right, but it's not harmful to set the sync round and call `wait-for-block-after` in this case since they should instantly succeed and also ensure the node is in a good state.

conduit/plugins/importers/algod/algod_importer.go
```diff
@@ -80,7 +82,120 @@ func init() {
...
+		case stat.CatchpointAcquiredBlocks > 0:
+			algodImp.logger.Infof("catchup phase 3 of 4 (Acquired Blocks): %d / %d", stat.CatchpointAcquiredBlocks, stat.CatchpointTotalBlocks)
+		case stat.CatchpointVerifiedAccounts > 0:
+			algodImp.logger.Infof("catchup phase 2 of 4 (Verified Accounts): %d / %d", stat.CatchpointVerifiedAccounts, stat.CatchpointTotalAccounts)
+		case stat.CatchpointProcessedAccounts > 0:
+			algodImp.logger.Infof("catchup phase 1 of 4 (Processed Accounts): %d / %d", stat.CatchpointProcessedAccounts, stat.CatchpointTotalAccounts)
+		default:
+			// Todo: We should expose catchupService.VerifiedBlocks via the NodeStatusResponse
+			algodImp.logger.Infof("catchup phase 4 of 4 (Verified Blocks)")
+		}
```
winder, line 149: How does this logging look? I think it changed a little with the KvMods.

conduit/plugins/importers/algod/algod_importer_test.go
```diff
@@ -69,20 +72,134 @@ func TestInitSuccess(t *testing.T) {
...
+					models.NodeStatus{LastRound: 1236},
+				}),
+				MakeStatusResponder("/v2/catchup/", 200, "")),
+			"",
+			[]string{
+				"catchup phase 1 of 4 (Processed Accounts): 1 / 1",
+				"catchup phase 2 of 4 (Verified Accounts): 1 / 1",
+				"catchup phase 3 of 4 (Acquired Blocks): 1 / 1",
+				"catchup phase 4 of 4 (Verified Blocks)",
+			}},
```
winder, line 164: nice

conduit/plugins/importers/algod/algod_importer_test.go
```diff
@@ -69,20 +72,134 @@ func TestInitSuccess(t *testing.T) {
...
+netaddr: %s
+catchup-config:
+  catchpoint: %s
+`, "follower", ttest.algodServer.URL, ttest.catchpoint)
+			_, err := testImporter.Init(ctx, conduit.MakePipelineInitProvider(&pRound, nil), plugins.MakePluginConfig(cfgStr), testLogger)
+			if ttest.err != "" {
+				assert.ErrorContains(t, err, ttest.err)
+			} else {
+				assert.NoError(t, err)
+			}
```
winder, line 189: Is there any point checking logs if there was an unexpected error condition?
```suggestion
			if ttest.err != "" {
				require.ErrorContains(t, err, ttest.err)
			} else {
				require.NoError(t, err)
			}
```

conduit/plugins/importers/filereader/filereader_test.go
```diff
@@ -3,6 +3,7 @@ package fileimporter
 import (
 	"context"
 	"fmt"
+	"github.com/algorand/conduit/conduit"
```
winder, line 6: nit: import order. I guess the linter doesn't check tests??

conduit/plugins/importers/algod/algod_importer_config.go
```diff
@@ -12,4 +12,14 @@ type Config struct {
...
+	CatchupConfig CatchupParams `yaml:"catchup-config"`
+}
+
+// CatchupParams provides information required to sync a follower node to the pipeline round
+type CatchupParams struct {
+	// <code>catchpoint</code> is the catchpoint used to run fast catchup on startup when your node is behind the current pipeline round.
+	Catchpoint string `yaml:"catchpoint"`
+	// <code>admin-token</code> is the algod admin API token.
+	AdminToken string `yaml:"admin-token"`
 }
```
winder, line 25: Could you add these to the sample config also?

conduit/plugins/importers/algod/algod_importer.go
Eric-Warehime: Looks like this
```
{"__type":"Conduit","_name":"main","level":"info","msg":"Starting Pipeline Initialization","time":"2023-03-27T07:19:54-07:00"}
{"__type":"Conduit","_name":"main","level":"info","msg":"Overriding default next round from 25101261 to 25101001.","time":"2023-03-27T07:19:54-07:00"}
{"__type":"importer","_name":"algod","level":"info","msg":"catchup phase 3 of 4 (Acquired Blocks): 7 / 1321","time":"2023-03-27T07:19:59-07:00"}
{"__type":"importer","_name":"algod","level":"info","msg":"catchup phase 3 of 4 (Acquired Blocks): 33 / 1321","time":"2023-03-27T07:20:04-07:00"}
{"__type":"importer","_name":"algod","level":"info","msg":"catchup phase 3 of 4 (Acquired Blocks): 62 / 1321","time":"2023-03-27T07:20:09-07:00"}
```
...
```
{"__type":"importer","_name":"algod","level":"info","msg":"Catchpoint catchup finished in 1m55.0583125s","time":"2023-03-27T07:21:49-07:00"}
```

conduit/plugins/metadata.go
```diff
@@ -1,5 +1,20 @@
 package plugins
 
+// Metadata returns fields relevant to identification and description of plugins.
+type Metadata struct {
+	Name         string
+	Description  string
+	Deprecated   bool
+	SampleConfig string
+	Type         PluginType
```
winder, line 9: I don't see where `Type` is used. Looks like it was previously unused as part of `PluginMetadata` and now it's still unused as part of `Metadata`?

conduit/pipeline/pipeline_test.go
```diff
@@ -900,18 +827,23 @@ type errorImporter struct {
...
 	Name:         "error_importer",
 	Description:  "An importer that errors out whenever GetBlock() is called",
 	Deprecated:   false,
 	SampleConfig: "",
 }
 
-func (e *errorImporter) Metadata() conduit.Metadata {
+func (e *errorImporter) Name() string {
+	//TODO implement me
+	panic("implement me")
```
shiqizng, line 839: forgot to change this?

shiqizng (commented):
just one question. everything else looks fine.

conduit/plugins/importers/algod/sample.yaml
```diff
@@ -13,3 +13,9 @@
 
     # Algod API token.
     token: ""
+
+    # Algod catchpoint catchup arguments
+    catchup-config:
+        catchpoint: ""
```
winder, line 19:
```suggestion
        # The catchpoint to use when running fast catchup. Select an appropriate catchpoint for your deployment.
        # They are published in the following locations:
        # mainnet: https://algorand-catchpoints.s3.us-east-2.amazonaws.com/consolidated/mainnet_catchpoints.txt
        # betanet: https://algorand-catchpoints.s3.us-east-2.amazonaws.com/consolidated/betanet_catchpoints.txt
        # testnet: https://algorand-catchpoints.s3.us-east-2.amazonaws.com/consolidated/testnet_catchpoints.txt
        catchpoint: ""
```

winder (approved):

PR #47, opened 2023-03-30
init: Option to write config file to stdout.

winder:
## Summary

While creating a docker container, I found that it could be used like `docker run algorand/conduit init` to create a config file. The problem is that output is written to a file.

This changes the behavior of `conduit init`. Previously it would create a directory named `data` containing the config file, now it writes the config file to stdout. To get the previous behavior you would use the command `conduit init -d data`.

## Test Plan

Unit tests are updated to cover the new behavior.

### DIFF
--- a/cmd/conduit/internal/initialize/init.go
+++ b/cmd/conduit/internal/initialize/init.go
@@ -2,7 +2,9 @@ package initialize
 
 import (
 	_ "embed"
+	"errors"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
@@ -20,6 +22,8 @@ import (
 var InitCommand = makeInitCmd()
 
 const defaultDataDirectory = "data"
+
+var StdoutAndPathErr = errors.New("do not provide a path and toStdout")
 
 //go:embed conduit.yml.example
 var sampleConfig string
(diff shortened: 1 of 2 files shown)


algochoi (approved):
Looking good to me and works on my local machine👍  
Minor question below

cmd/conduit/internal/initialize/init.go
```diff
@@ -41,25 +45,34 @@ func formatArrayObject(obj string) string {
...
-	}
+	var location string
 
-	configFilePath := filepath.Join(path, conduit.DefaultConfigName)
-	f, err := os.Create(configFilePath)
-	if err != nil {
-		return fmt.Errorf("runConduitInit(): failed to create %s", configFilePath)
+	// to get here, the path must be initialized
+	if configWriter == nil {
+		if path == "" && configWriter == nil {
```
shiqizng, line 57: configWriter == nil check is redundant on this line since it's already one in the previous line.

cmd/conduit/internal/initialize/init.go
```diff
@@ -106,19 +119,21 @@ func runConduitInit(path string, importerFlag string, processorsFlag []string, e
...
 
-	fmt.Printf("A data directory has been created %s.\n", location)
-	fmt.Printf("\nBefore it can be used, the config file needs to be updated with\n")
-	fmt.Printf("values for the selected import, export and processor modules. For example,\n")
-	fmt.Printf("if the default algod importer was used, set the address/token and the block-dir\n")
-	fmt.Printf("path where Conduit should write the block files.\n")
-	fmt.Printf("\nOnce the config file is updated, start Conduit with:\n")
-	fmt.Printf("  ./conduit -d %s\n", path)
+	// If a data dir is created, print some additional help.
+	if path != "" {
```
shiqizng, line 128: nit: it may be better to use `configWriter==nil` check here; one fewer var to track.  `path` is getting updated at line 58 and it only happens when `configWriter == nil`.

shiqizng (approved):
LGTM. minor suggestions.

cmd/conduit/internal/initialize/init.go
```diff
@@ -21,6 +23,8 @@ var InitCommand = makeInitCmd()
 
 const defaultDataDirectory = "data"
 
+var errStdoutAndPath = errors.New("do not provide a path and toStdout")
```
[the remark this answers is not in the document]
winder: Thanks, this made sense in an earlier iteration and I missed the error message

cmd/conduit/internal/initialize/init.go
```diff
@@ -41,25 +45,34 @@ func formatArrayObject(obj string) string {
...
 
-func runConduitInit(path string, importerFlag string, processorsFlag []string, exporterFlag string) error {
-	var location string
-	if path == "" {
-		path = defaultDataDirectory
-		location = "in the current working directory"
-	} else {
-		location = fmt.Sprintf("at '%s'", path)
+func runConduitInit(path string, configWriter io.Writer, importerFlag string, processorsFlag []string, exporterFlag string) error {
+	if configWriter != nil && path != "" {
```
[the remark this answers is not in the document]
winder: No, it's impossible for the user to hit this line with the current code. This code should probably be refactored to work more like `verifyFile` which calls `verifyYaml` to avoid the edge case. In lieu of that, I added the error check in case we use this function differently in the future.

cmd/conduit/internal/initialize/init.go
winder: configWriter is set to the file on line 74. I agree that the logic is strange, maybe I should just refactor to avoid the strange logic.

winder:
@shiqizng @algochoi I updated the code to simplify the things you pointed out.

PR #52, opened 2023-04-04
pipeline: RoundRequest plugin hook and postgres round request.

winder:
## Summary

New optional `RoundRequestor` interface that plugins may implement to request a specific round override during startup. This interface is implemented by the postgresql exporter to request a start round based on the database state.

Hook interface:
```
// RoundRequestor is an optional interface. Plugins should implement it if
// they would like to request an override to the pipeline managed round.
type RoundRequestor interface {
	// RoundRequest is called by the Conduit framework prior to `Init`.
	// This allows plugins with specific round requirements to override
	// the automatic round tracking provided by the Conduit pipeline.
	//
	// There are a few cases where this can be problematic. In these cases
	// an error will be generated and Conduit will fail to launch:
	// 1. Conduit is launched with --next-round-override
	// 2. Multiple plugins request a round override and the requested
	//    rounds are different.
	//
	// If it becomes common for multiple round overrides to be provided,
	// we will add an option to select which override to use. For example,
	// the override rule could be "command-line", "oldest", or the name
	// of a plugin which should be preferred.
	RoundRequest(config plugins.PluginConfig) (uint64, error)
}
```
## Test Plan

New unit tests.
Existing integration/unit tests.
Manual e2e testing.


conduit/plugins/exporters/postgresql/postgresql_exporter.go
```diff
@@ -53,6 +54,46 @@ func (exp *postgresqlExporter) Metadata() plugins.Metadata {
 	return metadata
 }
 
+// RoundRequest connects to the database, queries the round, and closes the
+// connection. If there is a problem with the configuration, an error will be
+// returned in the Init phase and this function will return 0.
+func (exp *postgresqlExporter) RoundRequest(cfg plugins.PluginConfig) (uint64, error) {
```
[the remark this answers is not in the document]
winder: extracted the common code into a helper function.

conduit/pipeline/pipeline_test.go
```diff
@@ -770,44 +694,75 @@ func TestPipelineMetricsConfigs(t *testing.T) {
...
+	mImporter.rndOverride = 1
+	mProcessor.rndOverride = 10
+
+	err := pImpl.Init()
+	assert.ErrorIs(t, err, MakeErrOverrideConflict(1, 10, false))
+
+	mProcessor.rndOverride = 0
+	mExporter.rndOverride = 10
+	err = pImpl.Init()
+	assert.ErrorIs(t, err, MakeErrOverrideConflict(1, 10, false))
```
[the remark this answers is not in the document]
winder: I split all of these new tests into separate named subtests which do not depend on state from earlier tests.

You are correct that this test was previously depending on a side effect from the "first test" which set `mImporter.rndOverride = 1`. In the new code this is hopefully much clearer.

conduit/pipeline/pipeline_test.go
```diff
@@ -819,6 +774,34 @@ func TestRoundOverwrite(t *testing.T) {
...
 		assert.Equal(t, uint64(i), pImpl.pipelineMetadata.NextRound)
 	}
+	pImpl.cfg.ConduitArgs.NextRoundOverride = 0
+
+	// plugin overrides - importer
+	mImporter.rndOverride = 10
+	err = pImpl.Init()
+	assert.Nil(t, err)
+	assert.Equal(t, uint64(10), pImpl.pipelineMetadata.NextRound)
+	mImporter.rndOverride = 0
```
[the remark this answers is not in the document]
winder: Cleaned this up so that the tests don't depend on side effects (or need lines like this to reset the state).

winder:
I'm having some trouble getting the e2e tests to work. Converting to draft until testing is finished.

conduit/pipeline/pipeline.go
```diff
@@ -201,17 +225,95 @@ func (p *pipelineImpl) registerPluginMetricsCallbacks() {
 	}
 }
 
-func (p *pipelineImpl) makeConfig(pluginType, pluginName string, cfg []byte) (config plugins.PluginConfig) {
-	config.Config = string(cfg)
+func (p *pipelineImpl) makeConfig(cfg NameConfigPair, pluginType plugins.PluginType) (*log.Logger, plugins.PluginConfig, error) {
```
winder, line 228: Expanded to initialize the logger and marshal the config file in addition to creating the plugin's data directory.

conduit/plugins/exporters/postgresql/postgresql_exporter.go
```diff
@@ -53,33 +54,69 @@ func (exp *postgresqlExporter) Metadata() plugins.Metadata {
...
+	nullLogger := logrus.New()
+	nullLogger.Out = io.Discard // no logging
+
+	db, _, err := createIndexerDB(nullLogger, true, cfg)
+	if err != nil {
+		// Assume the error is related to an uninitialized DB.
+		// If it is something more serious, the failure will be detected during Init.
+		return 0, nil
+	}
```
winder, line 99: I was not sure of a better way to handle this. Basically you can't create a read-only IndexerDb object if the DB hasn't been initialized.

conduit/plugins/importers/algod/algod_importer.go
```diff
@@ -270,13 +270,13 @@ func (algodImp *algodImporter) getDelta(rnd uint64) (sdk.LedgerStateDelta, error
...
+	// Note: this uses lenient decoding. GetRaw and msgpack.Decode would allow strict decoding.
+	err := (*common.Client)(algodImp.aclient).GetRawMsgpack(algodImp.ctx, &delta, fmt.Sprintf("/v2/deltas/%d", rnd), params, nil)
 	if err != nil {
-		return delta, err
+		return sdk.LedgerStateDelta{}, err
 	}
 
-	err = msgpack.Decode(bytes, &delta)
-	return delta, err
+	return delta, nil
```
winder, line 279: This is the source of the error I was running into. The master branch of algod has a new field in `ledgercore.StateDelta` and the SDK wasn't able to parse it with the strict decoder.

conduit/plugins/importers/algod/algod_importer.go
```diff
@@ -321,9 +321,9 @@ func (algodImp *algodImporter) GetBlock(rnd uint64) (data.BlockData, error) {
 				delta, err = algodImp.getDelta(rnd)
 				if err != nil {
 					if status.LastRound < rnd {
-						err = fmt.Errorf("ledger state delta not found: node round (%d) is behind required round (%d), ensure follower node has its sync round set to the required round", status.LastRound, rnd)
+						err = fmt.Errorf("ledger state delta not found: node round (%d) is behind required round (%d), ensure follower node has its sync round set to the required round: %w", status.LastRound, rnd, err)
 					} else {
-						err = fmt.Errorf("ledger state delta not found: node round (%d), required round (%d): verify follower node configuration and ensure follower node has its sync round set to the required round, re-deploying the follower node may be necessary", status.LastRound, rnd)
+						err = fmt.Errorf("ledger state delta not found: node round (%d), required round (%d): verify follower node configuration and ensure follower node has its sync round set to the required round, re-deploying the follower node may be necessary: %w", status.LastRound, rnd, err)
 					}
```
winder, line 327: This is where the failure occurred. The actual error was related to strict messagepack decoding, but the error wasn't included. I didn't see a way to detect 404s properly, so adding the error string for now.

conduit/pipeline/pipeline.go
```diff
@@ -269,76 +359,80 @@ func (p *pipelineImpl) Init() error {
...
-	if err != nil {
-		return fmt.Errorf("Pipeline.Init(): could not serialize Importer.Args: %w", err)
-	}
-	genesis, err := (*p.importer).Init(p.ctx, *p.initProvider, p.makeConfig("importer", importerName, configs), importerLogger)
-	if err != nil {
-		return fmt.Errorf("Pipeline.Init(): could not initialize importer (%s): %w", importerName, err)
-	}
-	(*p.initProvider).SetGenesis(genesis)
+	// Initialize Importer
+	{
```
Eric-Warehime, line 383: Why does this need a scope?

conduit/pipeline/pipeline.go
```diff
@@ -28,6 +28,30 @@ import (
...
+		pluginOrCLI: pluginOrCLI,
+		other:       other,
+		cli:         cli,
+	}
+}
+
+// Error implements the error interface.
+func (e ErrOverrideConflict) Error() string {
+	if e.cli {
+		return fmt.Sprintf("inconsistent round overrides detected: command line (%d), plugins (%d)", e.pluginOrCLI, e.other)
```
Eric-Warehime, line 50: This seems oddly named. The field is `pluginOrCLI` but the error message just says `command line`.

conduit/pipeline/pipeline.go
winder: It's not strictly needed. In this case I think it's nice to have all this code in one function, but the variables leak all over the place. Adding in a scope means that there aren't as many variables floating around.

I didn't want to completely refactor because that will come when we do the data pipe-lining work.

conduit/pipeline/pipeline.go
winder: The cli field is supposed to indicate which one it is. Maybe these inputs can be strings instead of numbers, then it can even identify which plugin provided the conflicting round. :thinking:

winder: Updated, this should be less convoluted now.

Eric-Warehime (approved):
LGTM. Do we need any docs changes along with this--maybe in the pgsql exporter plugin docs talking about automatically detecting the next round in the db?

conduit/plugins/exporters/postgresql/postgresql_exporter.go
```diff
@@ -53,33 +54,69 @@ func (exp *postgresqlExporter) Metadata() plugins.Metadata {
...
+
+	rnd, err := db.GetNextRoundToAccount()
+	if err != nil && err != idb.ErrorNotInitialized {
```
bbroder-algo, line 102: it's not clear to me why you are swallowing this error

conduit/plugins/importers/algod/algod_importer.go
```diff
@@ -321,9 +321,9 @@ func (algodImp *algodImporter) GetBlock(rnd uint64) (data.BlockData, error) {
 				delta, err = algodImp.getDelta(rnd)
 				if err != nil {
 					if status.LastRound < rnd {
```
bbroder-algo, line 323: why not do this check right after getting the status?  i.e. what algodImp.mode is this not valid for

conduit/pipeline/pipeline.go
```diff
@@ -201,17 +201,97 @@ func (p *pipelineImpl) registerPluginMetricsCallbacks() {
...
+		if err != nil {
+			return 0, err
+		}
+		rnd, err := part.RoundRequest(config)
 		if err != nil {
-			p.logger.Errorf("Unable to create plugin data directory: %s", err)
-			config.DataDir = ""
+			return 0, err
 		}
+		if pluginOverride != 0 && rnd != 0 && rnd != pluginOverride {
```
bbroder-algo, line 277: it's good that it's ok if they are the same override.

conduit/plugins/importers/algod/algod_importer.go
winder: Now I'm wondering if this is even possible, since `StatusAfterBlock` should ensure the Conduit round is not ahead of algod.

winder:
Followup PR to update documentation: https://github.com/algorand/conduit/pull/54

PR #57, opened 2023-04-14
algod importer: Fix archival mode and update tests.

winder:
## Summary

When adding in the recent follow mode automation the algod import mode the archival configuration was not accounted for. A bug in the unit test which treated sync calls to be treated as no-ops instead of not found errors allowed the regression to be missed.

## Test Plan

Update archival mode unit tests to return a 404 when sync endpoints are called to properly reflect the behavior of an archival node.

### DIFF
--- a/conduit/plugins/importers/algod/algod_importer.go
+++ b/conduit/plugins/importers/algod/algod_importer.go
@@ -34,7 +34,7 @@ const (
 )
 
 const (
-	archivalMode = iota
+	archivalMode = iota + 1
 	followerMode
 )
 
@@ -63,11 +63,12 @@ var algodImporterMetadata = plugins.Metadata{
 }
 
 func (algodImp *algodImporter) OnComplete(input data.BlockData) error {
-	if algodImp.mode != followerMode {
-		return nil
+	if algodImp.mode == followerMode {
+		_, err := algodImp.aclient.SetSyncRound(input.Round() + 1).Do(algodImp.ctx)
+		return err
 	}
-	_, err := algodImp.aclient.SetSyncRound(input.Round() + 1).Do(algodImp.ctx)
-	return err
+
+	return nil
 }
 
 func (algodImp *algodImporter) Metadata() plugins.Metadata {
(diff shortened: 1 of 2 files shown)


conduit/plugins/importers/algod/algod_importer_test.go
```diff
@@ -289,30 +299,47 @@ netaddr: %s
...
+				BlockResponder,
+				BlockAfterResponder,
+				MakeSyncRoundResponder(http.StatusNotFound))},
+		{
+			name: "archival",
+			mode: "archival",
+			algodServer: NewAlgodServer(GenesisResponder,
+				BlockResponder,
+				BlockAfterResponder,
+				MakeSyncRoundResponder(http.StatusNotFound))},
```
winder, line 318: Most of these `http.StatusOK` to `http.StatusNotFound` changes led to tests that fail without the bugfix.

conduit/plugins/importers/algod/algod_importer.go
```diff
@@ -63,11 +63,12 @@ var algodImporterMetadata = plugins.Metadata{
 }
 
 func (algodImp *algodImporter) OnComplete(input data.BlockData) error {
-	if algodImp.mode != followerMode {
-		return nil
+	if algodImp.mode == followerMode {
```
algochoi, line 66: These are non-functional changes correct? Why were the existing unit tests passing for archival nodes before and now return a 404?

tzaffi (commented):
Verified the new tests fail on old and pass with new. Left a minor question.

conduit/plugins/importers/algod/algod_importer.go
```diff
@@ -34,7 +34,7 @@ const (
 )
 
 const (
-	archivalMode = iota
+	archivalMode = iota + 1
```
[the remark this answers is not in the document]
winder: There's code that maps "" -> archivalMode, "archival" -> archivalMode and "follower" -> followerMode. I can remove this change, it doesn't effect anything.

conduit/plugins/importers/algod/algod_importer.go
winder: This change is non functional. The one below is the real fix.

conduit/plugins/importers/algod/algod_importer_test.go
```diff
@@ -345,15 +356,18 @@ func TestGetBlockContextCancelled(t *testing.T) {
...
 		{"follower", NewAlgodServer(GenesisResponder,
 			BlockResponder,
 			BlockAfterResponder, LedgerStateDeltaResponder,
 			MakeSyncRoundResponder(http.StatusOK))},
 	}
 
 	for _, ttest := range tests {
+		ttest := ttest
 		t.Run(ttest.name, func(t *testing.T) {
+			// this didn't work...
```
algochoi, line 369: Is test parallelization a blocker for PRs in this repo? It looks like it's not enforced in the repo by a linter [unlike go-algorand](https://github.com/algorand/go-algorand/blob/master/.golangci.yml#L20), but maybe it is implicitly done or a future todo.

algochoi (commented):
Looks good, one question about parallel tests

conduit/plugins/importers/algod/algod_importer_test.go
winder: It is not required but maybe it should be. I added a few on in this PR since I was changing the tests already.

algochoi (approved):

tzaffi (approved):
LGTM

PR #50, opened 2023-03-31
docker: Add docker to the release pipeline

winder:
## Summary

Add a docker multi-arch build + deployment configuration.

~For now the docker hub deployment is disabled.~ **Hyperflow has added credentials to the repo so that the container can  be deployed.**

Goreleaser provides binaries from the build step to the Dockerfile to ensure containers have the same files as the archives.

## Test Plan

Tested with `goreleaser release --skip-publish --snapshot --clean`

The files are here:
```
~$ docker images|grep conduit
algorand/conduit   1.0.1-next-amd64   0863fc121046   About a minute ago   96.9MB
algorand/conduit   latest-amd64       0863fc121046   About a minute ago   96.9MB
algorand/conduit   1.0.1-next-arm64   907990ee4310   About a minute ago   90.5MB
algorand/conduit   latest-arm64       907990ee4310   About a minute ago   90.5MB
```

The architecture appears as expected:
```
~$ docker image inspect algorand/conduit:latest-arm64 | grep Architecture
        "Architecture": "arm64",
~$ docker image inspect algorand/conduit:latest-amd64 | grep Architecture
        "Architecture": "amd64",
```

PR #58, opened 2023-04-20
docker: data dir & user permissions, install ca-certificates, better data dir location, more.

winder:
## Summary

This adds a number of small improvements over the first pass container:
* Enable the docker release.
* Better data dir / user permissions, [similar to #5276](https://github.com/algorand/go-algorand/pull/5276).
* Move data dir from `/conduit/data` to `/data`.
* Install `ca-certificates` to allow using an algod provider service.

## Testing

Similar to the previous docker PR: https://github.com/algorand/conduit/pull/50

Additional testing:

Mount data directory:
```
docker run -it --rm -v $(pwd)/local_data_dir:/data algorand/conduit:latest-snapshot-amd64
```

Check `CMD` overrides:
```
docker run -it --rm -v $(pwd)/local_data_dir:/data algorand/conduit:latest-snapshot-amd64 -h
docker run -it --rm -v $(pwd)/local_data_dir:/data algorand/conduit:latest-snapshot-amd64 init
docker run -it --rm -v $(pwd)/local_data_dir:/data algorand/conduit:latest-snapshot-amd64 list
```

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -30,11 +30,17 @@ jobs:
       - name: Setup Docker Buildx (docker multi-arch dependency)
         uses: docker/setup-buildx-action@v2
 
+      - name: dockerhub-login
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
       - uses: goreleaser/goreleaser-action@v4
         with:
           distribution: goreleaser
           version: latest
-          args: release --clean --skip-docker
+          args: release --clean
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
 # This dockerfile is used by goreleaser
 FROM debian:bullseye-slim
 
-RUN groupadd --gid=999 --system conduit && \
-    useradd --uid=999 --no-log-init --create-home --system --gid conduit conduit && \
+RUN groupadd --gid=999 --system algorand && \
+    useradd --uid=999 --no-log-init --create-home --system --gid algorand algorand && \
     mkdir -p /data && \
-    chown -R conduit.conduit /data && \
+    chown -R algorand.algorand /data && \
     apt-get update && \
     apt-get install -y gosu ca-certificates && \
     update-ca-certificates && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,12 @@ FROM debian:bullseye-slim
 RUN groupadd --gid=999 --system conduit && \
     useradd --uid=999 --no-log-init --create-home --system --gid conduit conduit && \
     mkdir -p /data && \
-    chown -R conduit.conduit /data
+    chown -R conduit.conduit /data && \
+    apt-get update && \
+    apt-get install -y ca-certificates && \
+    update-ca-certificates && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
 
 # binary is passed into the build
 COPY conduit /usr/local/bin/
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,14 +6,14 @@ RUN groupadd --gid=999 --system conduit && \
     mkdir -p /data && \
     chown -R conduit.conduit /data && \
     apt-get update && \
-    apt-get install -y ca-certificates && \
+    apt-get install -y gosu ca-certificates && \
     update-ca-certificates && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
 # binary is passed into the build
-COPY conduit /usr/local/bin/
-COPY docker/docker-entrypoint.sh /usr/local/bin/
+COPY conduit /usr/local/bin/conduit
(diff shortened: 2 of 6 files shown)


.github/workflows/goreleaser.yml
```diff
@@ -30,11 +30,17 @@ jobs:
       - name: Setup Docker Buildx (docker multi-arch dependency)
         uses: docker/setup-buildx-action@v2
 
+      - name: dockerhub-login
```
algochoi, line 33: Does this deploy to docker hub every time a release tag is created? I'm not familiar with the frequency that this workflow is triggered.

algochoi:
This works locally on my machine 👍

.github/workflows/goreleaser.yml
winder: Yes, it's been triggered once so far.

Dockerfile
```diff
@@ -1,14 +1,21 @@
 # This dockerfile is used by goreleaser
 FROM debian:bullseye-slim
 
-RUN useradd conduit
-RUN mkdir -p /conduit/data && \
-    chown -R conduit.conduit /conduit
+RUN groupadd --gid=999 --system algorand && \
```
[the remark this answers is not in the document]
winder: Added some comments.

docs/Docker.md
```diff
@@ -45,5 +43,17 @@ docker run algorand/conduit init --importer algod --processors filter_processor
 With `conduit.yml` in your current working directory,
 launch the container:
 ```
-docker run -it -v $(pwd)/conduit.yml:/conduit/data/conduit.yml algorand/conduit
+docker run -it -v $(pwd)/conduit.yml:/data/conduit.yml algorand/conduit
 ```
+
+# Mounting the Data Directory
```
tzaffi, line 49: Maybe switch the this section `# Mounting the Data Directory` with the previous `## Run with conduit.yml`? That's because the previous command is mounting and the later section explains mounting.

docs/Docker.md
```diff
@@ -45,5 +43,17 @@ docker run algorand/conduit init --importer algod --processors filter_processor
...
+
+The data directory is located at `/algod/data`. Mounting a volume at that location will allow you to resume the deployment from another container.
+
+## Volume Permissions
+
+The container executes in the context of the `algorand` user with UID=999 and GID=999 which is handled differently depending on your operating system or deployment platform. During startup the container temporarily runs as root in order to modify the permissions of /data. It then changes to the `algorand` user. This can sometimes cause problems, for example if your deployment platform doesn't allow containers to run as the root user.
```
tzaffi, line 55:
```suggestion
The container executes in the context of the `algorand` user with UID=999 and GID=999 which is handled differently depending on your operating system or deployment platform. During startup the container temporarily runs as root in order to modify the permissions of `/data`. It then changes to the `algorand` user. This can sometimes cause problems, for example if your deployment platform doesn't allow containers to run as the root user.
```

docs/Docker.md
winder: I was thinking that Mounting the data directory is more involved when it comes to permissions compared to only mounting `conduit.yml`. Maybe both sections should be combined somehow.

tzaffi: You can just refer to the next section by saying something like
```
...in your current working directory, launch the container. Notice that we're mounting the `conduit.yml` in our current working directory to `data/conduit.yml` in the container (see the next section):
```

tzaffi: Now that I think more about it, isn't it preferred to mount the directories as opposed to the config file?

tzaffi (commented):
WOMM but some outstanding non-functional discussion points

docs/Docker.md
```diff
@@ -45,5 +43,21 @@ docker run algorand/conduit init --importer algod --processors filter_processor
...
+docker run -it -v $(pwd)/local_data_dir:/data algorand/conduit
 ```
+
+## Volume Permissions
+
+The container executes in the context of the `algorand` user with UID=999 and GID=999 which is handled differently depending on your operating system or deployment platform. During startup the container temporarily runs as root in order to modify the permissions of `/data`. It then changes to the `algorand` user. This can sometimes cause problems, for example if your deployment platform doesn't allow containers to run as the root user.
+
+### Use specific UID and GID
+
+On the host system, ensure the directory being mounted uses UID=999 and GID=999. If the directory already has these permissions you may override the default user with `-u 999:999`.
```
bbroder-algo, line 63: Why would I want to override the default user with 999:999 if it already has those permissions?

winder: If you don't want the container to start as the root user. Added a clarifying comment.

docs/Docker.md
```diff
@@ -42,8 +40,25 @@ docker run algorand/conduit init --importer algod --processors filter_processor
 
 ## Run with conduit.yml
 
-With `conduit.yml` in your current working directory,
-launch the container:
+With `conduit.yml` in your current working directory, it can be mounted directoy to `/data/conduit.yml`. This is good for testing and some deployments which override the starting round. For a more complete deployment see the next section for how to mount the entire data directory.
```
tzaffi, line 43:
```suggestion
With `conduit.yml` in your current working directory, it can be mounted directly to `/data/conduit.yml`. This is good for testing and some deployments which override the starting round. For a more complete deployment see the next section which explains how to mount the entire data directory.
```

tzaffi (approved):
LGTM

PR #59, opened 2023-04-21
docs: Update plugin documentation.

winder:
## Summary

Create a README that embeds the sample config. This is still not perfect but I think it's an improvement.

+ Documentation is now complete. All configuration and features are described.
+ The documentation lives next to the code.
- The sample.yaml files are hand written and could diverge from the config.
- Markdown links cannot be used between tutorials and plugins. They would be different in GitHub and on the developer portal.

Followup Tasks:

* While it is nice to have the markdown files next to the code, it might be better to keep it in the `docs/plugins/{type}/` directories for the benefit of linking between documents.
 
* The `conduit-docs` tool could be used to generate the `sample.yaml` files, this would close the loop and be the best of both worlds.

## Test Plan

I'll manually deploy documentation after merging this.


.github/workflows/docs-pr.yml
```diff
@@ -39,6 +39,36 @@ jobs:
           # Update hand-written documentation
           rm -rf docs/docs/get-details/conduit/
           cp -r conduit/docs docs/docs/get-details/conduit
+
+          # Copy the plugin docs using this ChatGPT special.
+          # Set the directory to search
+          search_dir="conduit/plugins"
+          target_base_dir=docs/docs/get-details/conduit/plugins
```
shiqizng, line 46: is this copying READMEs from `conduit/plugins` to `docs`? what is this path `docs/docs/get-details/conduit/plugins`?

conduit/plugins/exporters/filewriter/README.md
```diff
@@ -55,12 +53,32 @@ You must use the proper expression type for the field your tag identifies based
...
+
+    # FilenamePattern is the format used to write block files. It uses go
+    # string formatting and should accept one number for the round.
+    # If the file has a '.gz' extension, blocks will be gzipped.
+    # Default: "%[1]d_block.json"
+    filename-pattern: "%[1]d_block.json"
+
+    # DropCertificate is used to remove the vote certificate from the block data before writing files.
+    drop-certificate: true
+```
```
shiqizng, line 79: this is a file_writer sample config but this readme is about transaction filtering?

.github/workflows/docs-pr.yml
winder: It's in the docs repo: https://github.com/algorand/docs/tree/staging/docs/get-details/conduit/plugins

conduit/plugins/exporters/filewriter/README.md
winder: Thank you, fixed!

shiqizng (approved):

PR #61, opened 2023-04-23
refactoring: Cleanup dependencies, port indexer utils, delete dead code.

winder:
## Summary

Cleanup code prior to breaking out plugin dependencies:
* Port indexer utility functions to avoid dependency.
* Delete dead code to remove indexer / ValidatedBlock dependency.
* Move simple type definitions to data package to avoid circular dependencies.

## Test Plan

N/A - non functional changes.


winder:
> LGTM, but why didn't the codecov run show up? Do we not have codecov enabled for conduit?

@Eric-Warehime Thats strange, I wonder why the job passes. This is in the log:
> [2023-04-23T14:21:30.491Z] ['error'] There was an error running the uploader: Error uploading to https://codecov.io: Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Commit sha does not match Circle build. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}

Eric-Warehime (approved):
Looking at the most recent CI runs, it seems like the codecov upload has been broken for a while. You don't need to fix that in this PR so I'm happy to approve as is and we can separately try to get that working again.

tzaffi (commented):
Looks good. My only nit/question is why should conduit's config live in the `data` package? I can see an argument that `config` is a kind of data. Also, creating a new package just for it is inconvenient and we want to avoid packages bloat.

tzaffi:
> > Looks good. My only nit/question is why should conduit's config live in the `data` package? I can see an argument that `config` is a kind of data. Also, creating a new package just for it is inconvenient and we want to avoid packages bloat.
> 
> @tzaffi I agree with most of this...

thx for clarifying

PR #62, opened 2023-04-25
ci: Migrate CircleCI automation to github actions. 

winder:
## Summary

Migrate CircleCI automation to github actions.

## Test Plan

Verify that this PR trigger the new automation, and the automation runs correctly.


docs/tutorials/IndexerMigration.md
```diff
@@ -235,7 +235,7 @@ Because our Conduit pipeline will use the Follower node's state delta API, we no
 volume. It can be removed.
 
 ### Step 3: Refactor our Indexer Writer to Conduit
-You're free to capture any data you like using Conduit. I'd encourage you to take a look at the [filter processor](./FilterDeepDive.md) and see
+You're free to capture any data you like using Conduit. I'd encourage you to take a look at the `filter processor` and see
```
bbroder-algo, line 238: I?

winder: Oops, wrong base branch.

bbroder-algo (approved):

tzaffi:
Do you want to remove `.circleci/config.yml` ?

winder:
I'll need to update the project's CI requirements after this is merged.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,54 @@
+name: Tests
```
tzaffi, line 1: Dropping the nightly tests?
image
Seems like they were identical to the on-commit tests (if they were even being run) so it's fine by me. If in the future we need to revive them, we can.

winder: They weren't being run.

The difference compared to the on-commit tests are:
```diff
- CHANNEL=stable
+ CHANNEL=nightly
```

I tried to make the workflow parameterized in this PR but it wasn't working. Agreed about bringing them back in the future. This is the same problem we have in SDKs: what version of algod should we be testing against?

winder: cc: @cce we've talked about this before, you might be curious. This is one way we pin versions between repositories. It means there is a gap where we don't know if the downstream tools are still compatible with algod.

PR #68, opened 2023-05-03
DevOps: Add CODEOWNERS file to restrict workflow editing

onetechnical:
## Summary

To restrict workflow edits, add a CODEOWNERS file.

## Test Plan

N/A


winder:
This change doesn't work because we're currently committing against master and the automation only triggers for the develop branch.

algobarb:
> This change doesn't work because we're currently committing against master and the automation only triggers for the develop branch.

This prevents someone else outside of Algorand to make PRs against this branch and write github actions files.

PR #72, opened 2023-05-07
enhancement: disallow Register() name dupes

tzaffi:
## Summary

Proposing that when `Register()` is called twice for the same plugin name, it will panic.
An alternative would involve changing the signature to return an `error`.

### Explanation

This is mostly about familiarizing myself with the Conduit repo. This is not an urgent issue.

As I was playing around with conduit [plugin templates](https://github.com/algorand/conduit-plugin-template) I realized there is a slight issue with plugin registration. 
> It's possible to register a plugin with the same name as another, and overwrite the first registration

The current functionality keeps a map from a user provided name to the actual plugin. If a plugin with the same name as another is registed a second time, it simply overwrites the first. This certainly isn't a pressing concern as we have complete control over what to register and what not to. 

## Test Plan

CI

### DIFF
--- a/conduit/plugins/exporters/exporter_factory_test.go
+++ b/conduit/plugins/exporters/exporter_factory_test.go
@@ -1,6 +1,7 @@
 package exporters
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/sirupsen/logrus"
@@ -46,4 +47,18 @@ func TestExporterByNameNotFound(t *testing.T) {
 	_, err := ExporterBuilderByName("barfoo")
 	expectedErr := "no Exporter Constructor for barfoo"
 	assert.EqualError(t, err, expectedErr)
+}
+
+// TestRegister verifies that the Register functions works as expected.
+func TestRegister(t *testing.T) {
+	mockName := "____mock"
+	assert.NotContains(t, Exporters, mockName)
+
+	Register(mockName, &mockExporterConstructor{})
+	assert.Contains(t, Exporters, mockName)
+
+	panicMsg := fmt.Sprintf("exporter %s already registered", mockName)
+	assert.PanicsWithError(t, panicMsg, func() {
+		Register(mockName, &mockExporterConstructor{})
+	})
 }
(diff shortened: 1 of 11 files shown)


conduit/plugins/exporters/exporter_factory_test.go
```diff
@@ -47,3 +48,17 @@ func TestExporterByNameNotFound(t *testing.T) {
 	expectedErr := "no Exporter Constructor for barfoo"
 	assert.EqualError(t, err, expectedErr)
 }
+
+// TestRegister verifies that the Register functions works as expected.
```
tzaffi, line 52:
```suggestion
// TestRegister verifies that Register works as expected.
```

conduit/plugins/importers/importer_factory_test.go
```diff
@@ -0,0 +1,53 @@
...
+func (i *mockImporter) Close() error                                { return nil }
+func (i *mockImporter) GetBlock(rnd uint64) (data.BlockData, error) { return data.BlockData{}, nil }
+
+type mockImporterConstructor struct{}
+
+func (c *mockImporterConstructor) New() Importer {
+	return &mockImporter{}
+}
+
+// TestRegister verifies that the Register functions works as expected.
```
tzaffi, line 41:
```suggestion
// TestRegister verifies that Register works as expected.
```

conduit/plugins/processors/processor_factory_test.go
```diff
@@ -52,3 +53,17 @@ func TestProcessorBuilderByNameNotFound(t *testing.T) {
 	expectedErr := "no Processor Constructor for barfoo"
 	assert.EqualError(t, err, expectedErr)
 }
+
+// TestRegister verifies that the Register functions works as expected.
```
tzaffi, line 57:
```suggestion
// TestRegister verifies that Register works as expected.
```

pkg/cli/cli.go
```diff
@@ -19,9 +19,12 @@ import (
...
 
 var (
-	logger     *log.Logger
-	conduitCmd = MakeConduitCmd()
+	logger *log.Logger
 
+	// ConduitCmd is the root command for conduit
+	ConduitCmd = MakeConduitCmd()
+
+	//Banner is TUI banner for conduit
```
tzaffi, line 27:
```suggestion
	//Banner is the banner for conduit's pipeline
```

winder (approved):

winder:
Good idea, nice catch.

Issue #51, opened 2023-04-03
Add optional Telemetry tracking for Conduit


algoanne:
We would like to make it possible to enable Telemetry tracking on Conduit deployments. This would be entirely optional (an option in the Conduit config file), but can be valuable both for debugging and understanding usage patterns. 

Do NOT include any PII, such as IP address, or specifics of which exact filters are set if using a filter processor (for example). 

For this first ticket, we will just gather 3 pieces of data. 
- At startup, report the generated ID & the current round. We will add data in follow-up tickets. 
- add 1 error, e.g. "pipeline is dead"

Stick with the ELK stack (Elastic Search, logstash, Kibana - a pattern we're familiar with). 

**Out of scope**
The following sets of information would be useful: 
- error reporting 
- hardware setup 
- conduit-specific setup (i.e. which plugins users are using)

Kibana dashboard

PR #74, opened 2023-05-15
telemetry: Push telemetry event to OpenSearch on pipeline startup

algochoi:
## Summary

Sends telemetry events with a GUID upon pipeline startup. Adds a telemetry config to `conduit.yml` and saves GUID to the pipeline metadata file. Telemetry files are in `telemetry/` and added a README for a quick summary.

More context & research in this gist: https://gist.github.com/algochoi/5fe7381100f2af8cf8c6b0e828be1e56

https://github.com/algorand/conduit/issues/51

## Test Plan

Unit tests for telemetry configs during pipeline initialization.

Tested on local OpenSearch cluster and Algorand dev cluster. See the gist above on how to replicate this.


conduit/data/block_export_data.go
```diff
@@ -16,6 +18,8 @@ type InitProvider interface {
 	GetGenesis() *sdk.Genesis
 	SetGenesis(*sdk.Genesis)
 	NextDBRound() sdk.Round
+	GetTelemetryClient() *telemetry.OpenSearchClient
+	SetTelemetryClient(*telemetry.OpenSearchClient)
```
winder, line 22: I think `Set` can be a property of the constructor / implementation rather than included as part of the interface.

algochoi: Good point, removed from interface in b4acb9aa974320d5743aab8c7bd68eb30c1e9b44

conduit/pipeline/pipeline.go
```diff
@@ -254,8 +274,26 @@ func (p *pipelineImpl) Init() error {
...
+		var telemetryErr error
+		telemetryClient, telemetryErr = p.initializeTelemetry()
+		if telemetryErr != nil {
+			p.logger.Warn("Telemetry initialization failed. Continuing without telemetry.")
+		}
+		// Try sending a startup event. If it fails, log a warning and continue
+		event := telemetryClient.MakeTelemetryStartupEvent()
+		if telemetryErr = telemetryClient.SendEvent(event); telemetryErr != nil {
+			p.logger.Warnf("failed to send telemetry event: %s", telemetryErr)
+		}
```
winder, line 292: What do you think about wrapping the OpenSearch client in our own object that implements SendEvent?

It could manage the logging and allow us to swap in a different client in the future if it became useful. Since this is part of the public interface, it could be problematic to swap out or upgrade the client in the future.

conduit/telemetry/telemetryCommon.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// Client represents the Telemetry client and config
+type Client interface {
+	MakeTelemetryStartupEvent() Event
+	SendEvent(event Event) error
+}
+
+// OpenSearchClient holds the OpenSearch client and TelemetryConfig
+type OpenSearchClient struct {
```
winder, line 38: Oh! It already is wrapped. Could you rename this to TelemetryClient?

conduit/telemetry/README.md
```diff
@@ -0,0 +1,16 @@
+# Conduit Telemetry
+
+The goal is to gain information about how our users use Conduit in the wild. This currently does not include things like logging (logrus) or metrics (Prometheus and Datadog for internal performance tests).
+
+We want to closely follow how algod reports telemetry. However, there is some divergence in the client, e.g. currently using ElasticSearch, but in process of migrating to OpenSearch. We currently use the official [Go opensearch client](https://opensearch.org/docs/latest/clients/go/).
```
winder, line 5: I'd suggest leaving out the comment about algod. Maybe just mention that the structure is similar to algod and that we're using opensearch, that way it wont become outdated once algod migrates.

conduit/telemetry/telemetryCommon.go
```diff
@@ -0,0 +1,41 @@
...
+	URI       string
+	Name      string
+	GUID      string
+	Index     string
+	UserName  string
+	Password  string
+}
+
+// Event represents a single event to be emitted to OpenSearch
+type Event struct {
```
winder, line 22: Could you add the version string?

pkg/cli/internal/initialize/conduit.test.init.default.yml
```diff
@@ -36,4 +36,11 @@ exporter:
...
 
-
+# Enable telemetry for conduit
+telemetry:
+  enabled: true
+  # These configs define default values for initializing the telemetry server.
+  uri: ""
+  index: ""
+  username: ""
+  password: ""
```
winder, line 46:
```suggestion
  # By default the following fields will be configured to send data to Algorand.
  # Do store your own telemetry events, they can be overridden.
  # uri: ""
  # index: ""
  # username: ""
  # password: ""
```

pkg/cli/internal/initialize/conduit.yml.example
```diff
@@ -33,3 +33,12 @@ processors:
...
 %s
+
+# Enable telemetry for conduit
+telemetry:
+  enabled: true
+  # These configs define default values for initializing the telemetry server.
+  uri: ""
+  index: ""
+  username: ""
+  password: ""
```
winder, line 44:
```suggestion

  # By default the following fields will be configured to send data to Algorand.
  # To store your own telemetry events, they can be overridden.
  # uri: ""
  # index: ""
  # username: ""
  # password: ""
```

conduit/telemetry/README.md
```diff
@@ -0,0 +1,16 @@
...
+## Implementation
+
+Telemetry related functionality will be put into the `conduit/telemetry` directory.
+
+### Configurations
+The conduit.yml will have a new telemetry boolean. If true, `telemetry.Config` (which hold the OpenSearch URI, hardcoded credentials, GUID, Index name) will be initialized. The GUID is generated when the pipeline is started, and saved to the pipeline's `metadata.json`. If this field already exists, it will be read from the file. 
+The index name in algod is set to the Chain ID (e.g. mainnet and testnet). In conduit, we consolidate to a common index name for now.
+
+### Client
+The configuration and client is represented by the `telemetry.Client` interface and will use the `OpenSearchClient` by default. Users should be able to define their own clients to send telemetry events. This will also have client functions to write to OpenSearch using the official Go client. There is no plan to use a particular logging library as of now, and just marshalls the `telemetry.Event` struct into JSON before sending the event.
```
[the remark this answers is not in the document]
algochoi: oops you're right, I will update this file

conduit/telemetry/telemetryCommon.go
algochoi: I could, but the current lint configurations output an error because of stuttering:

```
exported: type name will be used as telemetry.TelemetryClient by other packages, and that stutters; consider calling this Client (revive)
    type TelemetryClient struct {
```

Should I disable this check and rename to `TelemetryClient` anyway? I'm okay either way.

conduit/telemetry/telemetryCommon.go
algochoi: good point, added conduit LongVersion() string to the event and fixed existing tests: [234ca36](https://github.com/algorand/conduit/pull/74/commits/234ca3690cf169c537c9d9f10b90c430fd728a52)

conduit/telemetry/README.md
algochoi: Agreed, deleted this section [30adcc3](https://github.com/algorand/conduit/pull/74/commits/30adcc329f7bc80046c681c1fdec7f31509d6f7d)

conduit/telemetry/telemetryCommon.go
algochoi: > I always forget about this. telemetry.Client seems good too, how about just Client?

The interface is named `Client` so there would be a collision unfortunately :(

pkg/cli/internal/initialize/conduit.yml.example
algochoi: Updated in [96a7c9c](https://github.com/algorand/conduit/pull/74/commits/96a7c9cfa4b3858828e10fa74910d2492fc53654)

conduit/telemetry/telemetryCommon.go
winder: In that case outside of this package you can reference the interface instead of `telemetry.OpenSearchClient`

conduit/pipeline/pipeline.go
```diff
@@ -195,6 +196,25 @@ func (p *pipelineImpl) pluginRoundOverride() (uint64, error) {
 	return pluginOverride, nil
 }
 
+// initializeTelemetry initializes telemetry and reads or sets the GUID in the metadata.
+func (p *pipelineImpl) initializeTelemetry() (*telemetry.OpenSearchClient, error) {
```
winder, line 200: Places like this:
```suggestion
func (p *pipelineImpl) initializeTelemetry() (*telemetry.Client, error) {
```

conduit/telemetry/telemetryCommon.go
algochoi: Good point! Will do this

conduit/data/block_export_data.go
```diff
@@ -16,6 +18,7 @@ type InitProvider interface {
 	GetGenesis() *sdk.Genesis
 	SetGenesis(*sdk.Genesis)
 	NextDBRound() sdk.Round
+	GetTelemetryClient() telemetry.Client
```
winder, line 21: The other nice thing about this approach is that you can create a test client for unit testing the event.

conduit/pipeline/pipeline.go
```diff
@@ -257,8 +277,26 @@ func (p *pipelineImpl) Init() error {
...
+		// If telemetry cannot be initialized, log a warning and continue
+		// pipeline initialization.
+		var telemetryErr error
+		telemetryClient, telemetryErr = p.initializeTelemetry()
+		if telemetryErr != nil {
+			p.logger.Warn("Telemetry initialization failed. Continuing without telemetry.")
+		}
+		// Try sending a startup event. If it fails, log a warning and continue
+		event := telemetryClient.MakeTelemetryStartupEvent()
+		if telemetryErr = telemetryClient.SendEvent(event); telemetryErr != nil {
```
algochoi, line 293: Currently, without config modifications, we'll always log a warning here.

winder: Why?

algochoi: The pipeline tries to initialize the client [based on the config file](https://github.com/algorand/conduit/pull/74/files/61f253077c38c6408854525e524bc88595ef6a2c#diff-71752f54b1d598e2902b12d3d654ca6fd80b11c9848da2390967c3664c173b69R204). Since the default config file doesn't specify a URI, we don't send an event for now. We can change the default config.yml when we have an OpenSearch production server ready or alternatively hardcode the URI/credentials in like we do in algod.

winder: Is it the `Telemetry initialization failed` warning, or `failed to send telemetry event`?

Let's default `Enabled` to false, so this block wouldn't be hit anyway.

algochoi: > Is it the Telemetry initialization failed warning, or failed to send telemetry event?

Currently you'll get both warnings - it might be too redundant so I'll change it to send one or the other.

> Let's default Enabled to false, so this block wouldn't be hit anyway.

Good point, will do 👍

winder: Try to come up with a design that lets us gracefully disable telemetry. Keep in mind that plugins are going to be grabbing the telemetryClient object and they wont be able to check for Enabled.

algochoi: > Try to come up with a design that lets us gracefully disable telemetry. Keep in mind that plugins are going to be grabbing the telemetryClient object and they wont be able to check for Enabled.

The telemetryClient object has its own `Enabled` bool: https://github.com/algorand/conduit/pull/74/files#diff-ada99fa4b32a319d368310f322561a3e2292fefcd48533e763dff3069fe64cf0R11

This reads the config file upon pipeline initialization, checks if the config file has set the telemetry enabled bool to true, and initializes the client object (which then sets the client's bool to true). I think the current implementation lets us dynamically disable telemetry on the client as well.

winder (approved):

conduit/telemetry/telemetry.go
```diff
@@ -0,0 +1,84 @@
...
+	"context"
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/opensearch-project/opensearch-go"
+	"github.com/opensearch-project/opensearch-go/opensearchapi"
```
tzaffi, line 14: 👍

conduit/telemetry/telemetry.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/algorand/conduit/version"
+)
+
+// MakeTelemetryConfig initializes a new TelemetryConfig.
+func MakeTelemetryConfig(telemetryURI, index, username, password string) Config {
+	return Config{
+		Enable:   true,
+		URI:      telemetryURI,
+		GUID:     uuid.New().String(), // Use Google UUID instead of go-algorand utils
```
tzaffi, line 24: It seems like an uncontroversial decision to me, but the comment got me curious about the pros/cons.

conduit/telemetry/telemetry.go
```diff
@@ -0,0 +1,84 @@
...
+
+// initializeOpenSearchClient creates a new OpenSearch client.
+func initializeOpenSearchClient(cfg Config) (*opensearch.Client, error) {
+	client, err := opensearch.NewClient(opensearch.Config{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		},
+		Addresses: []string{cfg.URI},
+		// These credentials are here intentionally. Not a bug.
+		Username: cfg.UserName,
```
tzaffi, line 39: Not for this PR, but worth a discussion. We might want to leverage a library such as https://github.com/zalando/go-keyring to that users can have a way to interact with secrets stored more securely on their system than directly in the config file.

go.mod
```diff
@@ -6,7 +6,9 @@ require (
 	github.com/algorand/go-algorand-sdk/v2 v2.0.0-20230531123858-3a0efd6a6398
 	github.com/algorand/go-codec/codec v1.1.8
 	github.com/algorand/indexer v0.0.0-20230601214318-cb2f42818f82
+	github.com/google/uuid v1.3.0
 	github.com/jackc/pgx/v4 v4.13.0
+	github.com/opensearch-project/opensearch-go v1.1.0
```
tzaffi, line 11: Are you stuck with 1.1 because of our version of go?

tzaffi (approved):
This is great. No we have the ability to start assessing what's happening in the wild. We ought to follow up with a couple of more `SendEvents()` for some crucial information such as succeeding/failing to catchup.

I left a couple of questions and nits, but nothing blocking.

conduit/telemetry/README.md
```diff
@@ -0,0 +1,13 @@
+# Conduit Telemetry
+
+The goal is to gain information about how our users use Conduit in the wild. This currently does not include things like logging (logrus) or metrics (Prometheus and Datadog for internal performance tests).
+
+## Implementation
+
+Telemetry related utilities are in this directory.
+
+### Configurations
+The conduit.yml will have a new telemetry boolean, and optional fields to enter the OpenSearch URI, credentials (username/password), and Index name. If true, `telemetry.Config` will be initialized with the optional fields. The GUID is generated when the pipeline is started, and saved to the pipeline's `metadata.json`. If this field already exists, it will be read from the file.
```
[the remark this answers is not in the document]
algochoi: Good catch, fixed in [f32193f](https://github.com/algorand/conduit/pull/74/commits/f32193fff02e1d0d45e4d8df5da0592879df8163)

go.mod
algochoi: Good question, I initially chose it for the widest compatibility ([matrix](https://opensearch.org/docs/latest/tools/index/)), but it's now unclear why we really need it. I upgraded to use the v2 client in [f32193f](https://github.com/algorand/conduit/pull/74/commits/f32193fff02e1d0d45e4d8df5da0592879df8163).

conduit/telemetry/telemetry.go
algochoi: It's a good point - we also do this in go-algorand: https://github.com/algorand/go-algorand/blob/master/logging/telemetryConfig.go#L37

We hardcode the telemetry credentials because it's write-only anyway. It's certainly weird if a user has to do this for their custom telemetry client though. We also discussed using API tokens which might be a little better instead of hardcoded passwords as well.

Issue #75, opened 2023-05-15
Fix `lint` in PR Process


algochoi:
## Problem

Our PR workflow is not enforcing `make lint` right now. There is a [reviewdog action](https://github.com/algorand/conduit/blob/master/.github/workflows/reviewdog.yml#L20) that checks for lint errors, but does not fail even when running `make lint` locally would fail.

It looks like there is a panic in the process: https://github.com/algorand/conduit/actions/runs/4983728162/jobs/8921142359#step:6:82 

Running the action locally seems to trigger lint errors, e.g. `act pull_request -j lint`

PR #76, opened 2023-05-15
lint: Fix the `lint` github workflow

algochoi:
## Summary

Fix reviewdog linter in the github PR workflow. 

Before this, [reviewdog silently caused a panic](https://github.com/algorand/conduit/actions/runs/4983728162/jobs/8921142359#step:6:82), but the workflow succeeded anyway. This change pins the reviewdog step to explicitly use Go 1.17.13 (uses latest 1.20.* version by default, which was causing a panic) and sets to check with `nofilter` (only checks `added` lines of code by default). 

## Test Plan

Locally run (requires [act](https://github.com/nektos/act)): `act pull_request -j lint`

PR `lint` workflow fails when there is a golangci lint error: https://github.com/algorand/conduit/actions/runs/4984363520/jobs/8922595805?pr=76

Closes https://github.com/algorand/conduit/issues/75

### DIFF
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -27,3 +27,4 @@ jobs:
           tool_name: "Lint Errors"
           level: "error"
           fail_on_error: true
+          filter_mode: "nofilter"
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -24,5 +24,5 @@ jobs:
           golangci_lint_flags: "-c .golangci.yml --allow-parallel-runners"
           reporter: "github-pr-review"
           tool_name: "Lint Errors"
-          level: "error"
+          level: "warn"
           fail_on_error: true
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -22,7 +22,8 @@ jobs:
         with:
           golangci_lint_version: "v1.47.3"
           golangci_lint_flags: "-c .golangci.yml --allow-parallel-runners"
+          go_version: "1.17"
           reporter: "github-pr-review"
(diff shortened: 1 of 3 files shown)


tzaffi:
Any objections to adding a check to ensure that `go mod tidy` doesn't make any changes?

@algorand/lamprey

Eric-Warehime:
> Any objections to adding a check to ensure that `go mod tidy` doesn't make any changes?
> 
> @algorand/lamprey

In my opinion that's going to cause a lot of failures for unrelated PRs and be pretty noisy.

.github/workflows/reviewdog.yml
```diff
@@ -24,5 +24,5 @@ jobs:
           golangci_lint_flags: "-c .golangci.yml --allow-parallel-runners"
           reporter: "github-pr-review"
           tool_name: "Lint Errors"
-          level: "error"
```
Eric-Warehime, line 27: Changing the log level on CI lint will make it so that lint warnings cause failure correct? It's fine if we want that to be the standard--I just want to make sure I understand.

algochoi: Sorry, that was just for testing (and likely not relevant) - there is currently a panic caused in the Github action and I'm in the process of debugging this

tzaffi:
> > Any objections to adding a check to ensure that `go mod tidy` doesn't make any changes?
> > @algorand/lamprey
> 
> In my opinion that's going to cause a lot of failures for unrelated PRs and be pretty noisy.

It's not clear to me why this would be a problem with unrelated PR's. Let's discuss next standup.

tzaffi (approved):
Great improvement, thank you!

Eric-Warehime (approved):

PR #78, opened 2023-05-15
Bug-Fix: don't attempt to get deltas at round 0

tzaffi:
## Summary
Don't attempt to query deltas from a follower node for round 0. Instead, just declare that we need to catch up and log:
> No state deltas are ever available for round 0

## Test Plan

CI - introduced new unit test cases

### DIFF
--- a/conduit/plugins/importers/algod/algod_importer.go
+++ b/conduit/plugins/importers/algod/algod_importer.go
@@ -230,6 +230,10 @@ func checkRounds(logger *logrus.Logger, catchpointRound, nodeRound, targetRound 
 
 func (algodImp *algodImporter) needsCatchup(targetRound uint64) bool {
 	if algodImp.mode == followerMode {
+		if targetRound == 0 {
+			algodImp.logger.Info("No state deltas are ever available for round 0")
+			return true
+		}
 		// If we are in follower mode, check if the round delta is available.
 		_, err := algodImp.getDelta(targetRound)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


shiqizng (approved):

bbroder-algo:
What does the algod state delta endpoint do if you request the zero round?

tzaffi:
> What does the algod state delta endpoint do if you request the zero round?

Current error stack:
> "msg":"Unable to fetch state delta for round 0: HTTP 404: {\"message\":\"failed retrieving State Delta

conduit/plugins/importers/algod/algod_importer.go
```diff
@@ -230,6 +230,10 @@ func checkRounds(logger *logrus.Logger, catchpointRound, nodeRound, targetRound
 
 func (algodImp *algodImporter) needsCatchup(targetRound uint64) bool {
 	if algodImp.mode == followerMode {
+		if targetRound == 0 {
+			algodImp.logger.Info("No state deltas are ever available for round 0")
+			return true
+		}
```
winder, line 236: Why does this return true? If the target round is 0 I think it would go ahead and no-op in the catchup function, but is it needed? Maybe in the case where Conduit is fully caught up and it is told to go back to 0?

I think this would be slightly more correct so that it still checks for the block.
```
	if algodImp.mode == followerMode && targetRound != 0 {
```

Or maybe even
```
if targetRound == 0 {
    targetRound++
}
```

winder: cc: @Eric-Warehime

tzaffi: There were only two choices of what to return and that seemed the most reasonable. But we could change the API,.

PR #83, opened 2023-05-28
enhancement: plugin interface

tzaffi:
# Re-organize Plugin Interfaces

These changes grew out a personal project to create a binary conduit plugin (or plugins) which will help me calculate taxes. As I go along and spot various opportunites for streamlining, clarifying the requirements, and docs, I'll continue putting up small<sup>[[1]](#footnote1)</sup> PR's of this kind.

Except for the last commit (see **Odd Finding** below), all changes are non-functional. I'm expecting to roll back the last commit, but it's worth further discussion.

## Summary of Changes

* Introducing `.markdownlint.yml`. This isn't enforced in CI but while using the VS Code extension it helped me identify some bugs in our docs. For example, some [improperly formatted list items](https://developer.algorand.org/docs/get-details/conduit/Development/#creating-a-plugin).
* Introducing `Plugin interface` extending `PluginMetadata` and with methods `Config()` and `Close()` in `conduit/plugins/plugin.go`. These are extended by the `Importer`, `Processor` and `Exporter` interfaces instead of `PlugingMetadata` resulting in changes in:
  * `conduit/plugins/exporters/exporter.go`
  * `conduit/plugins/importers/importer.go`
  * `conduit/plugins/processors/processor.go`
*  Small edits in the following docs:
   * `docs/Configuration.md`
   * `docs/GettingStarted.md`
   * `docs/tutorials/WritingBlocksToFile.md`
   * `docs/tutorials/IndexerMigration.md`
* Add checks that the `genesis` value returned from newly introduced `Importer.GetGenesis()` meets expectations in various unit tests.

### Odd Finding
Notice that [all tests pass](https://github.com/algorand/conduit/actions/runs/5114020152/jobs/9193845564?pr=83) after I [commented out all usages](https://github.com/algorand/conduit/pull/83/commits/155e5ecae57476647cece74192ef40a9990a5dd1) of `Config()`. Is this actually used anywhere? If not, we should consider removing it from the plugin interfaces.

#### Upshot of Odd Finding
After a discussion, we decided to streamline the `Plugin` interface as follows:

- [x] Remove `Config()` from the interfaces
- [x] Flatten `PluginMetadata` directly into `Plugin` interface
- [x] Add `Init()` to `Plugin` interface and...
  - [x] ...split out `GetGenesis()` from `Importer.Init()`

---

<a name="footnote1">[1]</a>: It was still _small_ at the original time that I created this PR description.

PR #87, opened 2023-05-31
misc: Docs Update + Other non-functional changes

tzaffi:
I pulled out the doc updates from draft PR #83 and also fixed a couple of scoping issues in our tests.

## Summary of Changes

* Introducing `.markdownlint.yml`. This isn't enforced in CI but while using the VS Code extension it helped me identify some bugs in our docs. For example, some [improperly formatted list items](https://developer.algorand.org/docs/get-details/conduit/Development/#creating-a-plugin).
*  Small edits in the following docs:
   * `docs/Configuration.md`
   * `docs/Development.md`
   * `docs/GettingStarted.md`
   * `docs/tutorials/WritingBlocksToFile.md`
* loop variable capture fixes in:
   * `conduit/pipeline/pipeline_test.go`
   * `conduit/plugins/importers/algod/algod_importer_test.go` (also tightening the test expectations around the genesis object received from `Init()`)

## TODO

- [x] Shall we add `exportloopref` to our linters?

```.golangci.yml
linters:
  enable:
    - exportloopref
```


conduit/pipeline/pipeline_test.go
```diff
@@ -612,6 +612,7 @@ func TestRoundRequestError(t *testing.T) {
 func TestRoundOverride(t *testing.T) {
 	// cli override NextRound, 0 is a test for no override.
 	for i := 0; i < 10; i++ {
+		i := i
```
algochoi, line 615: good catch

.markdownlint.yml
```diff
@@ -0,0 +1,7 @@
+default: true
```
algochoi, line 1: I personally use VSCode so I am fine with this change but wondering what others think about adding IDE/editor/extension specific config files in the root dir.

algochoi (approved):
Looks like good changes to me!

.markdownlint.yml
tzaffi: What do you think about adding `exportloopref` ?

conduit/pipeline/pipeline_test.go
tzaffi: see cbe1692504021f71c69af1b636aa00cbe6015285

FYI - it didn't actually catch the problem. But we're also using it in `go-algorand` so it should be safe.

.markdownlint.yml
tzaffi: see cbe1692504021f71c69af1b636aa00cbe6015285

FYI - it didn't actually catch the problem. But we're also using it in `go-algorand` so it should be safe.

winder (approved):

PR #89, opened 2023-05-31
Fix catchup error on first run and add retry forever option.

winder:
## Summary

Two small changes:
1. The needs catchup logic was incorrectly reporting an error on first run because there was no catchpoint available for round 0.
2. Add handling for retry = 0 to cause the pipeline to keep retrying forever. This could be useful if algod is being upgraded or otherwise down for maintenance.

## Test Plan

Update unit tests.


tzaffi:
Is [this failing](https://github.com/algorand/conduit/actions/runs/5135460944/jobs/9240905709?pr=89#step:9:120) because the test needs updating or because the changes aren't correct?

conduit/pipeline/pipeline.go
```diff
@@ -383,12 +383,13 @@ func (p *pipelineImpl) Start() {
...
 		pipelineRun:
 			metrics.PipelineRetryCount.Observe(float64(retry))
-			if retry > p.cfg.RetryCount {
+			if retry > p.cfg.RetryCount && p.cfg.RetryCount != 0 {
 				p.logger.Errorf("Pipeline has exceeded maximum retry count (%d) - stopping...", p.cfg.RetryCount)
 				return
 			}
 
 			if retry > 0 {
+				p.logger.Infof("Retry number %d resuming after a retry delay.", retry)
```
tzaffi, line 392: ignorable nit: you could add the retry delay to the message

conduit/pipeline/pipeline_test.go
```diff
@@ -744,15 +767,39 @@ func TestPipelineRetryVariables(t *testing.T) {
...
+			assert.WithinDurationf(t, before.Add(testCase.totalDuration), after, epsilon, msg)
+			if testCase.retryCount == 0 {
+				assert.GreaterOrEqual(t, errImporter.GetBlockCount, uint64(1))
+			} else {
+				assert.Equal(t, errImporter.GetBlockCount, testCase.retryCount+1)
+			}
+			done = true
+			for hook := hook.AllEntries(); len(hook) > 0; hook = hook[1:] {
 
+			}
```
tzaffi, line 794:
```suggestion
```
This seems like a no-op. Am I missing something?

tzaffi (commented):
Looks good but I had one remaining question.

conduit/pipeline/pipeline_test.go
winder: woops, this was a co-pilot special that I forgot to remove

tzaffi: At least we're not a CRISPR shop and this isn't CRISPR Co-Pilot suggesting a human ending virus mutation.

tzaffi (approved):
LGTM

PR #98, opened 2023-06-12
docs: fix filter docs to replace exact.

winder:
## Summary

The equal and exact types were merged in https://github.com/algorand/indexer/pull/1468 but the documentation was not properly updated. This PR updates the documentation to replace `exact` with `equal`.

### DIFF
--- a/conduit/plugins/processors/filterprocessor/README.md
+++ b/conduit/plugins/processors/filterprocessor/README.md
@@ -40,13 +40,12 @@ Example:
 ### `expression-type`
 The expression type is a selection of one of the available methods for evaluating the expression. The current list of
 types is
-* `exact`: exact match for string values.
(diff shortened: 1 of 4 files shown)


shiqizng (approved):

tzaffi (approved):

algochoi (commented):
Looks good thanks! Should expression-type be wrapped in quotes? I only see it being defined like this in `conduit/plugins/processors/filterprocessor/README.md` (which is a markdown file).

winder:
> Looks good thanks! Should expression-type be wrapped in quotes? I only see it being defined like this in `conduit/plugins/processors/filterprocessor/README.md` (which is a markdown file).

added quotes.

algochoi (approved):

PR #110, opened 2023-06-23
Define and update e2e file in Makefile.

winder:
## Summary

Moves the CI_E2E_FILENAME definition into the Makefile so that `make e2e-conduit` is easier to run locally.

This PR also pins the filename to a more recent version.

The variable is defined with `?=` to allow it to be overridden by an environment variable.

## Test Plan

Existing CI.


algochoi (approved):
Nice, the E2E tests work on my machine after this PR!

PR #113, opened 2023-06-30
pipelineImpl: de-pointerize plugin interface fields

tzaffi:
### Non-breaking changes:
* Modifying the plugin fields in the `pipelineImpl` struct to be "concrete" interface types.

### Quasi-breaking changes:
* Rename `*Builder` variables and types to be `*Constructor`
* Rename `Constructor` interface which was used for constructing **Importer**s to be `ImporterConstructor` in line with other such plugin related types.

## Test Plan

CI

### Testing TODO's
- [x] run locally with the block generator and see that it still works

#### Results of the local block generator run
It's working just as before. Below is a screen shot comparing the run's report with what's in the database. Additionally, the _conduit-log_ had no errors.

image


conduit/pipeline/pipeline.go
```diff
@@ -306,11 +306,11 @@ func (p *pipelineImpl) Init() error {
 		if err != nil {
 			return fmt.Errorf("Pipeline.Init(): could not make %s config: %w", p.cfg.Importer.Name, err)
 		}
-		err = (*p.importer).Init(p.ctx, *p.initProvider, pluginConfig, importerLogger)
+		err = (p.importer).Init(p.ctx, *p.initProvider, pluginConfig, importerLogger)
```
winder, line 309: Are the parens still needed?
```suggestion
		err = p.importer.Init(p.ctx, *p.initProvider, pluginConfig, importerLogger)
```

conduit/pipeline/pipeline.go
```diff
@@ -431,8 +431,129 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
 	metrics.ImportedTxnsPerBlock.Observe(float64(len(block.Payset)) + float64(innerTxn))
 }
 
-// Start pushes block data through the pipeline
+// Start is what Start was before the beginning of this branch
```
tzaffi, line 434: OOPS - bad merge... I need to fix this

conduit/pipeline/pipeline.go
```diff
@@ -574,8 +696,7 @@ func MakePipeline(ctx context.Context, cfg *data.Config, logger *log.Logger) (Pi
 		return nil, fmt.Errorf("MakePipeline(): could not build importer '%s': %w", importerName, err)
 	}
 
-	importer := importerBuilder.New()
-	pipeline.importer = &importer
+	pipeline.importer = importerBuilder.New()
```
tzaffi, line 699: This change and the ones below in this file aren't needed, but seem ok.

conduit/pipeline/pipeline.go
```diff
@@ -431,8 +431,129 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
+				if err != nil {
+					p.logger.Errorf("%v", err)
+					p.setError(err)
+					// Retry logic
+					continue
+				}
+				fetchBlocks <- &blkData
+			}
+		}
+	}()
```
tzaffi, line 555:
```suggestion
```

conduit/pipeline/pipeline.go
```diff
@@ -431,8 +431,103 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
 	metrics.ImportedTxnsPerBlock.Observe(float64(len(block.Payset)) + float64(innerTxn))
 }
 
-// Start pushes block data through the pipeline
+// Start is what Start was before the beginning of this branch
```
tzaffi, line 434:
```suggestion
// Start pushes block data through the pipeline
```

conduit/pipeline/pipeline.go
tzaffi: f507ba58379a85661a19bdc51d6a2eb57ae40a78

conduit/pipeline/pipeline.go
tzaffi: Well, partially their needed, but we don't really need to remove the extra variable assignment as I'm doing here.

conduit/pipeline/pipeline.go
```diff
@@ -477,18 +477,18 @@ func (p *pipelineImpl) Start() {
 					start := time.Now()
 					for _, proc := range p.processors {
 						processorStart := time.Now()
-						blkData, err = (*proc).Process(blkData)
+						blkData, err = (proc).Process(blkData)
```
winder, line 480:
```suggestion
						blkData, err = proc.Process(blkData)
```

conduit/pipeline/pipeline.go
```diff
@@ -477,18 +477,18 @@ func (p *pipelineImpl) Start() {
...
 						if err != nil {
 							p.logger.Errorf("%v", err)
 							p.setError(err)
 							retry++
 							goto pipelineRun
 						}
-						metrics.ProcessorTimeSeconds.WithLabelValues((*proc).Metadata().Name).Observe(time.Since(processorStart).Seconds())
+						metrics.ProcessorTimeSeconds.WithLabelValues((proc).Metadata().Name).Observe(time.Since(processorStart).Seconds())
```
winder, line 487:
```suggestion
						metrics.ProcessorTimeSeconds.WithLabelValues(proc.Metadata().Name).Observe(time.Since(processorStart).Seconds())
```

conduit/pipeline/pipeline.go
winder: Not sure why this variable is named "Builder". This is the "Constructor" and "New" is the single function defined in the constructor interface. The function above should be `importers.ImporterConstructorByName`.

What you have here is correct, previously it had to be assigned separately so that we could get the pointer.

winder (approved):

conduit/pipeline/pipeline.go
winder: Thanks. Up to you, I don't feel very strongly. Was just calling out a possible reason for the discrepancy because you mentioned it in your top comment.

tzaffi: Went ahead and merged. We should discuss how to communicate this for our next release.

Issue #117, opened 2023-07-17
PostgreSQL exporter delete-task does not seem to be working 


PSjoe:
### The PostgreSQL exporter `delete-task` option does not seem to be working

We were hoping to use the `delete-task` option to keep our dev databases a bit smaller.  Our `conduit.yml` file looks like this:  
```
exporter:
  name: postgresql
  config:
    connection-string: <redacted>
      sslmode=disable
    delete-task:
      rounds: 1000
      interval: 20
```

However, when running against either a fully-synced PostgreSQL instance or an instance that is starting from 0, it does not appear that the `txn` table is being culled.  Using the above configs, I can see the following in the DB: 

```
betanet=> select round from txn;
 round 
-------
   357
 12166
 12178
 12188
 13011
 13018
 13020
 14810
 14817
 14823
 14864
 15655
 15661
 15668
 18075
 36998
(16 rows)
```

I'd have expected the early rows to have been thrown out as conduit moves through later blocks.

### Your environment

* Conduit 1.2.0 (37a13fe)
* PostgreSQL 13.10
* Algod 3.16.3.stable [rel/stable] (commit #cbccc6de)
* Ledger:  Betanet

### Steps to reproduce

1. Bootstrap an empty PosgreSQL DB 
2. Start Conduit using the PostgreSQL exporter from block 0 using the configs above.

### Expected behaviour
Only transactions within the last 1000 blocks would be present 

### Actual behaviour
All transactions appear to be present.  Table continues to get longer as conduit works its way through the ledger.

tzaffi:
Another useful data point for our investigation is how long the experiment ran for. If it was only a couple of seconds long, it may not have triggered a pruning of the database since that's on a timer that runs every 2 seconds.

PSjoe:
Config file is as follows: 
```
log-level: INFO

# If no log file is provided logs are written to stdout.
#log-file:

# Set to 0 to retry forever.
retry-count: 10

retry-delay: "1s"


# Whether or not to print the conduit banner on startup.
hide-banner: false

# When enabled prometheus metrics are available on '/metrics'
metrics:
    mode: ON
    addr: ":9999"
    prefix: "conduit"

# The importer is typically an algod follower node.
importer:
    name: algod
    config:
        mode: "follower"
        netaddr: http://localhost:48081
        token: <redacted>
        catchup-config:
            auto: true
            admin-token: <redacted>


# Zero or more processors may be defined to manipulate what data
# reaches the exporter.
processors:

[...]

PR #119, opened 2023-07-19
bugfix: postgresql exporter delete task

tzaffi:
## Summary

The delete task wasn't running because the exporter config is wasn't actually being set in the exporter struct.

This PR includes:
* The bugfix in `conduit/plugins/exporters/postgresql/postgresql_exporter.go` and `conduit/plugins/exporters/postgresql/util/prune.go`
* End-to-end test enhancement to verify the fix (`pyproject.toml` and `*.py`). This required providing the capibility to query the indexer database.
* Changing the CLI descriptions for the `conduit init` command to be more in line with the other such descriptions (`pkg/cli/internal/initialize/init.go`)
* Miscellaneous lint errors that I was experiencing locally, but that _don't_ seem to be happening in CI.
* Fixing `TestMetrics` in `conduit/pipeline/pipeline_test.go` which started failing mysteriously.

## TODO

- [ ] Close #121 deleting its branch
- [ ] Shouldn't the debug statements added actually be traces?

## Issues

#117 

## Test Plan

A new E2E test is introduced to ensure that the delete task works as expected. I tried running this new test against the original code, and it failed in the expected manner. 

See: #121


conduit/data/config.go
```diff
@@ -7,7 +7,7 @@ import (
 	"time"
 
 	log "github.com/sirupsen/logrus"
-	"gopkg.in/yaml.v3"
+	yaml "gopkg.in/yaml.v3"
```
tzaffi, line 10: It was failing my local `golangci-lint` but not entirely sure why it's ok in CI. Many other similar changes.

conduit/pipeline/pipeline_test.go
```diff
@@ -914,32 +915,32 @@ func TestMetrics(t *testing.T) {
 		if strings.HasSuffix(*stat.Name, metrics.BlockImportTimeName) {
 			found++
 			// 1 hour in seconds
-			assert.Contains(t, stat.String(), "sample_count:1 sample_sum:3600")
+			assert.Regexp(t, regexp.MustCompile("sample_count:1 +sample_sum:3600"), stat.String())
```
tzaffi, line 918: @winder 
For mysterious reasons, this tests started failing on me and I had to be more lenient on the assertions in order to pass.

conduit/plugins/exporters/postgresql/postgresql_exporter.go
```diff
@@ -54,11 +54,12 @@ func (exp *postgresqlExporter) Metadata() plugins.Metadata {
...
 
 // createIndexerDB common code for creating the IndexerDb instance.
-func createIndexerDB(logger *logrus.Logger, readonly bool, cfg plugins.PluginConfig) (idb.IndexerDb, chan struct{}, error) {
+func createIndexerDB(logger *logrus.Logger, readonly bool, cfg plugins.PluginConfig) (idb.IndexerDb, chan struct{}, ExporterConfig, error) {
 	var eCfg ExporterConfig
 	if err := cfg.UnmarshalConfig(&eCfg); err != nil {
-		return nil, nil, fmt.Errorf("connect failure in unmarshalConfig: %v", err)
+		return nil, nil, eCfg, fmt.Errorf("connect failure in unmarshalConfig: %v", err)
 	}
+	logger.Debugf("createIndexerDB: eCfg.Delete=%+v", eCfg.Delete)
```
tzaffi, line 62: I can remove the debug statements if others opine in this direction.

e2e_tests/src/e2e_conduit/e2econduit.py
```diff
@@ -43,21 +44,30 @@ def main():
     else:
         logging.basicConfig(level=logging.INFO)
     sourcenet = args.source_net
-    source_is_tar = False
+    # source_is_tar = False
```
tzaffi, line 47:
```suggestion
```

e2e_tests/src/e2e_conduit/fixtures/importers/follower_algod.py
```diff
@@ -1,15 +1,11 @@
-import atexit
 import glob
```
tzaffi, line 1: changes due to applying black

e2e_tests/src/e2e_conduit/scenarios/follower_indexer_scenario.py
```diff
@@ -1,14 +1,108 @@
...
+class FollowerIndexerScenario(Scenario):
+    def __init__(self, sourcenet):
+        super().__init__(
+            name="follower_indexer_scenario",
+            importer=importers.FollowerAlgodImporter(sourcenet),
+            processors=[],
+            exporter=exporters.PostgresqlExporter(),
+        )
+
+    def get_validation_errors(self) -> list[str]:
```
tzaffi, line 17: @shiqizng 
This breaks my promise not to modify existing tests. I found it convenient during development to add this validation and then modify it for the delete task. It might be useful to keep, or it may cause an occasional flake. If anyone expresses the view that it should be removed, I'll gladly remove.

e2e_tests/src/e2e_conduit/scenarios/follower_indexer_scenario.py
```diff
@@ -1,14 +1,108 @@
...
+        """
+        validation checks that txn either contains no rows or that
+        the max round contained is the same as the network's lastblock
+        """
+
+        # sleep for 3 seconds to allow delete_task iteration to wake up after 2 seconds
+        # for its final pruning when the conduit is all caught up with the network
```
tzaffi, line 76:
```suggestion
        # for its final pruning when Conduit is all caught up with the network
```

pkg/cli/internal/initialize/init.go
```diff
@@ -146,9 +146,9 @@ func makeInitCmd() *cobra.Command {
 	var processors []string
 	cmd := &cobra.Command{
 		Use:   "init",
-		Short: "initializes a Conduit data directory",
-		Long: `Initializes a conduit.yml file and writes it to stdout. By default
-the config file uses an algod importer in follower mode and a block
+		Short: "Initializes a conduit configuration",
```
tzaffi, line 149: Changing the description in line with the other commands. If you try `conduit help` you'll see what I'm talking about.

tzaffi (commented):
suggestions

e2e_tests/src/e2e_conduit/scenarios/__init__.py
```diff
@@ -9,12 +9,20 @@
...
     """Data class for conduit E2E test pipelines"""
+
     name: str
     importer: ImporterPlugin
     processors: list[PluginFixture]
     exporter: PluginFixture
     accumulated_config: dict = field(default_factory=dict)
     conduit_dir: str = ""
 
+    def get_validation_errors(self) -> list[str]:
```
shiqizng, line 20: we actually run accounting validation in indexer e2e. we should probably create a ticket to migrate that to conduit.

e2e_tests/src/e2e_conduit/scenarios/follower_indexer_scenario.py
```diff
@@ -1,14 +1,108 @@
...
+
+        errors = []
+        try:
+            num_txn_rows = idb.get_table_row_count("txn")
+
+            if num_txn_rows > 0:
+                try:
+                    first_txn_round, last_txn_round = idb.get_txn_min_max_round()
+
+                    if first_txn_round != last_txn_round:
```
shiqizng, line 91: why should we expect first_txn_round==last_txn_round?

e2e_tests/src/e2e_conduit/scenarios/follower_indexer_scenario.py
```diff
@@ -1,14 +1,108 @@
...
+
+        idb = IndexerDB.from_connection_string(
+            self.exporter.config_input["connection-string"]
+        )
+
+        errors = []
+        try:
+            num_txn_rows = idb.get_table_row_count("txn")
+
+            if num_txn_rows > 0:
```
shiqizng, line 87: num_txn_rows==0 could potentially be an error. it could mean that no txns were imported when we expected some txns rows?

e2e_tests/src/e2e_conduit/scenarios/follower_indexer_scenario.py
tzaffi: Because the delete task is only supposed to keep one round at most.

e2e_tests/src/e2e_conduit/scenarios/follower_indexer_scenario.py
tzaffi: Yeah, that a good point. But hopefully the other test case without delete task would catch that.

algochoi (commented):
I didn't look closely at the tests but the lint + fix LGTM

e2e_tests/src/e2e_conduit/scenarios/__init__.py
tzaffi: Thanks for the suggestion. I created issue #123

conduit/pipeline/pipeline_test.go
winder: I would guess it's related to upgrading the prometheus versions.

Am I reading the regex correctly, that there is additional whitespace and some control character changes (quotes instead of angles?) but the data is otherwise the same?

tzaffi: Yep. We now have `{` instead of `<` and unpredictable white space.

conduit/plugins/exporters/postgresql/postgresql_exporter.go
```diff
@@ -112,10 +113,11 @@ func (exp *postgresqlExporter) Init(ctx context.Context, initProvider data.InitP
 	exp.ctx, exp.cf = context.WithCancel(ctx)
 	exp.logger = logger
 
-	db, ready, err := createIndexerDB(exp.logger, false, cfg)
+	db, ready, exporterCfg, err := createIndexerDB(exp.logger, false, cfg)
 	if err != nil {
 		return fmt.Errorf("db create error: %v", err)
 	}
+	exp.cfg = exporterCfg
```
winder, line 120: You could also put this right at the top and pass it into the createIndexerDB function:

```
	if err := cfg.UnmarshalConfig(&exp.cfg); err != nil {
		return fmt.Errorf("failure in unmarshalConfig: %v", err)
	}
	db, ready, err := createIndexerDB(exp.logger, false, exp.cfg)
```

alternatively, slight nit
```suggestion
	db, ready, exp.cfg, err := createIndexerDB(exp.logger, false, cfg)
	if err != nil {
		return fmt.Errorf("db create error: %v", err)
	}
```

e2e_tests/src/e2e_conduit/scenarios/follower_indexer_scenario.py
```diff
@@ -1,14 +1,108 @@
...
+                try:
+                    first_txn_round, last_txn_round = idb.get_txn_min_max_round()
+
+                    if first_txn_round != last_txn_round:
+                        errors.append(
+                            f"Indexer first txn round {first_txn_round} is not equal to last txn round {last_txn_round}"
+                        )
```
winder, line 94: Could you clarify these error messages? Rather than describing what the state is try to describe why the state is wrong.

winder (changes requested):
Fix look good.

The one thing I'd like to see changed is the error messages for the integration test. If the test were to fail, the current error messages are not much to work with.

conduit/plugins/exporters/postgresql/postgresql_exporter.go
tzaffi: I need to revert this change because otherwise, I need to change the assignment to a non-creating assignment and declare the variables `db, ready, err` on separate lines.

image

tzaffi: reverted in e69c231be21c9ed46e1f0f3b1afd1478d3cf4f43

e2e_tests/src/e2e_conduit/scenarios/follower_indexer_scenario.py
tzaffi: addressed in e69c231be21c9ed46e1f0f3b1afd1478d3cf4f43

tzaffi: Here's what the new error message looks like:
image

You can also [see this on CI](https://github.com/algorand/conduit/actions/runs/5617839792/job/15222532402?pr=121#step:11:136).

winder: Thank you!

winder (approved):

PR #125, opened 2023-07-21
Default to retrying after errors forever.

winder:
## Summary

On an errors, Indexer will retry forever. Users have expressed an appreciation for this behavior because they can perform maintenance on algod and it doesn't disrupt service on the downstream applications.

This changes Conduit's default retry on error behavior to have unlimited attempts, which matches how Indexer 2.x works.

### DIFF
--- a/pkg/cli/internal/initialize/conduit.yml.example
+++ b/pkg/cli/internal/initialize/conduit.yml.example
@@ -6,7 +6,7 @@ log-level: INFO
 
 # Number of retries to perform after a pipeline plugin error.
 # Set to 0 to retry forever.
-retry-count: 10
+retry-count: 0
 
 # Time duration to wait between retry attempts.
 retry-delay: "1s"


pkg/cli/internal/initialize/conduit.yml.example
```diff
@@ -6,7 +6,7 @@ log-level: INFO
 
 # Number of retries to perform after a pipeline plugin error.
 # Set to 0 to retry forever.
-retry-count: 10
+retry-count: 0
```
tzaffi, line 9: I'm personally against making this be the default. In a happy path setup for an experienced user this may very well make sense. But for a newbie starting out and with some configuration that's off the effect will be to have conduit hang instead of failing fast.

But I wasn't aware that this was the default for indexer 2.0 and others should express there thoughts.

winder: I was surprised too. The current default was made for pretty much the reason you're describing. Failing fast seems like a good thing. That said, several users brought up the behavior as something they appreciated about Indexer.

We don't retry during Init, so plugins have a chance to kill the app during startup if they want.

That said, I've considered adding some sort of retry for the importer to smooth over the case where Conduit starts before Algod.

Issue #118, opened 2023-07-18
Pipeline the Pipeline


tzaffi:
## Problem

Following up from #11. With the design agreed upon, implement the design.

## Solution

See the [Conduit Pipelining Design Doc](https://docs.google.com/document/d/1IdgHgIZ4JM-9xJNnx3ya142DFTFKMkoEgl-sTr4Dloc/edit).

## Dependencies

Draft PR: #114 

## Urgency

High

PR #97, opened 2023-06-12
enhancement: graceful pipeline interrupt

tzaffi:
## Summary

This continues [an original draft PR](https://github.com/tzaffi/conduit/pull/3), but this time against `algorand/conduit/master`.

Gracefully handling interrupt signal. The non-gracefulness was discovered during block-generator testing.  The following new printouts are being generated (which without the new go-routine are _NOT_ printed).

```sh
{"__type":"Conduit","_name":"main","level":"info","msg":"Pipeline received stopping signal <interrupt>, stopping pipeline. p.pipelineMetadata.NextRound: 14","time":"2023-06-09T14:47:06-05:00"}
{"__type":"importer","_name":"algod","level":"trace","msg":"importer algod.GetBlock() called BlockRaw(14) err: context canceled","time":"2023-06-09T14:47:06-05:00"}
{"__type":"importer","_name":"algod","level":"error","msg":"error getting block for round 14 (attempt 0): context canceled","time":"2023-06-09T14:47:06-05:00"}
{"__type":"importer","_name":"algod","level":"trace","msg":"importer algod.GetBlock() called StatusAfterBlock(13) err: context canceled","time":"2023-06-09T14:47:06-05:00"}
{"__type":"Conduit","_name":"main","level":"error","msg":"GetBlock ctx error: context canceled","time":"2023-06-09T14:47:06-05:00"}
{"__type":"Conduit","_name":"main","level":"info","msg":"Retry number 1 resuming after a 1s retry delay.","time":"2023-06-09T14:47:06-05:00"}
{"__type":"importer","_name":"algod","level":"info","msg":"importer algod.Close() at round 14","time":"2023-06-09T14:47:07-05:00"}
{"__type":"Conduit","_name":"main","level":"info","msg":"Pipeline.Stop(): Importer (algod) closed without error","time":"2023-06-09T14:47:07-05:00"}
{"__type":"importer","_name":"algod","level":"info","msg":"importer algod.Close() at round 14","time":"2023-06-09T14:47:07-05:00"}
{"__type":"Conduit","_name":"main","level":"info","msg":"Pipeline.Stop(): Importer (algod) closed without error","time":"2023-06-09T14:47:07-05:00"}
{"__type":"exporter","_name":"postgresql","level":"info","msg":"exporter postgresql.Close() at round 14","time":"2023-06-09T14:47:07-05:00"}
[...]

Issue #100, opened 2023-06-12
Graceful Pipeline Exit


tzaffi:
## Problem

Currently, when the main conduit process receives a shutdown signal, it abruptly tears down all its goroutines even though the pipeline shares a common `context.Context` object that is supposed to make possible to graceful in a more deterministic fashion. In particular, it is currently not possible to coordinate a new requirement such as having each plugin finish the current round, before exit.

## Solution

This story posits the following goals (though after further discussion we may want to trim the goals or break them out into seperate issues):

* Handle an interrupt signal in a way that won't cause duplicate calls to plugin `Stop()`. As a starting point refer to #97 but better crafting of where to place the signal handling is required.
* Consider adding a new `conduit stop` command that looks for the PID file and sends an interrupt to that process
* Consider the possibility of letting each plugin go to the end. Even better make this configurable via the [type of interrupt signal received](https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/syscall/zerrors_linux_amd64.go;l=1321). EG:
   * `conduit stop now` would send an `os.Interrupt` (or a `kill -TERM $(PROCNUM) ) and kills all plugin processes immediately
   * `conduit stop end-of-round` produces something like `kill -USR1 $(PROCNUM)` and kills plugin process in the following order at the end of each round: `Importer` &rarr; `Processors` &rarr; `Exporters`

## Dependencies

None

## Urgency

Low - I haven't heard of other developers in the communit complain of this issue.

winder:
Copying similar description from #99

### Subject of the issue

The pipeline object is managed using a context with a cancel function and a wait group. This setup is strange.

Rather than managing the process internal to the pipeline, `Start()` should be renamed to `Run()`, and `Run()` should be blocking. If the context is cancelled `Run()` can call `Stop()` automatically.

The caller would then use a waitgroup and run the pipeline in a go-routine if it wants to. `Run()` could also return the error directly rather than implementing `setError` and `Error`.

PR #128, opened 2023-07-24
Pipeline the Pipeline

tzaffi:
# Description

Allowing for moderate concurrency in the pipeline but without sacrificing its sequential integrity.

## Summary of Changes
* `conduit/pipeline/common.go`: introducing generic retrying of pipeline methods via `Retries()` and `RetriesNoOutput()`
* `conduit/pipeline/pipeline.go`: 
  * error/cancellation handling:
    * modify the pipeline's cancellation function to have a cause, and expose it via the `WhyStopped()` method
    * `joinError()` instead of `setError()` to the pipeline's error property
  * introducing goroutines for the importer, processors, exporters, and round launcher by refactoring `Start()` and introducing methods `ImportHandler()`, `ProcessorHandler()`, and `ExporterHandler()`
  * E2E end of test signal: modified the info-level log at the end of a round to look like **FINISHED Pipeline round: 110. UPDATED Pipeline round: 111** and added _WARNING_ commentary to be careful about changing this format as it would break the E2E test.
* `conduit/pipeline/pipeline_bench_test.go` - benchmarker for the pipeline that includes _sleeping plugins_ with an importer, 2 processors, and an exporter.
* `pkg/cli/cli.go`: remove a line break from the final error printout

## Issues

#118 

## TODO

- [x] Cleanup E2E code
- [x] Merge in #129 
- [x] Don't embed `PipelineData` inside of `BlockData`
- [x] Remove `OStart()` and any resulting orphans in `pipeline.go`
- [x] Merge latest from `master` which includes 30 second timeout #133
- [x] ~Should we send a telemetry observation based on `WhyStopped()` ?~ NOT FOR NOW. Reconsider this for #97 
- [x] ~What should be logged and how should `internal-tools/.../logstats.go` be modified?~ This is under the umbrella of the "Logging Plugin Performance" [thread](https://github.com/algorand/conduit/pull/128#discussion_r1284363465)
- [x] ~Should we send metrics for the callbacks?~ Not in this PR
- [x] Test on EC2 with the final changes in place.
- [x] Update the results from that test in the spreadsheet table image below.

### `pipeline_bench_test.go`
Running a new benchmark test twice on the original code and the new, we have the following results. Note the most pertinent results for the typical indexer DB population use case is `exporter_10ms_while_others_1ms`:

| Benchmark Name                                           | Original rounds/sec | Pipelining rounds/sec | Pipelining v Original (%) |
|----------------------------------------------------------|--------------|----------------|---------------------------|
| vanilla_2_procs_without_sleep-size-1-8                   | 3077         | 3309.5         | +7%                    |
| uniform_sleep_of_10ms-size-1-8                           | 22.32        | 79.815         | +250%                  |
| exporter_10ms_while_others_1ms-size-1-8                  | 63.405       | 78.565         | +24%                   |
| importer_10ms_while_others_1ms-size-1-8                  | 65.535       | 91.255         | +39%                   |
| first_processor_10ms_while_others_1ms-size-1-8           | 60.28        | 89.175         | +48%                   |

### Block Generator Results
Running the block generator test using `SCENARIO = scenarios/config.allmixed.small.yml` for 30s, with the original code and the new, each time for 2 experiments we have:

| Reset database?    | Original rounds/30 sec | Pipelining rounds/30 sec | Pipelining v Original (%) |
|--------------------|------------------------|--------------------------|---------------------------|
| Reset              | 301                    | 400                      | +33%                   |
| No Reset           | 295                    | 418                      | +41%                   |

### Local test network 5 minute sprint
I used the [Justfile command](https://gist.github.com/tzaffi/0b456532a7d6511ede7fe1bdae241e75#file-justfile-L471)

```sh
❯ just conduit-bootstrap-and-go 300
```
to bootstrap testnet and run a postgresql exporter against it for 300 seconds. I ran it a number of times against both the original pipeline and the new one. Here are the experimental results:

| Log Level | Reps | Original rounds/300 sec (logs/round) | Pipelining rounds/300 sec (logs/round) | Pipelining v Original (%) |
|-----------|------|-----|--------------------------|---------------------------|
| TRACE    | 3     | 3718  (7.0)  | 3509 (14.0)  | -5.6%  😢   |
| INFO     | 2     | 4578.5 (3.0) |  4423.5 (3.0)  | -3.4% 😢 |

### On EC2 - **CLASSIC** vs. **PIPELINING** vs. **30 Second Timeout** vs. **FINAL**

I ran catchup tests for 4 versions of conduit:
* **CLASSIC** - what was on `master` at the time of the run
* **PIPELINING** - the head of the `pipelining` branch at the same time
* **30 Second Timeout** - the `pipelining` branch after a 30 second timeout was introduced in the algod importer
* **FINAL** - commit https://github.com/algorand/conduit/pull/128/commits/867973f82e586f7d736e53e44eaf6d1f08871275 which is essentially the code meant for merging

There are much more detailed results in a google sheets document, but the summary is:

#### SUMMARY

image


conduit/pipeline/pipeline.go
```diff
@@ -431,8 +444,216 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
+				_, lastError = RetriesNoOutput(p.pipelineMetadata.encodeToFile, p.cfg.ConduitArgs.ConduitDataDir, p, "pipelineMetadata save")
+				if lastError != nil {
+					lastError = fmt.Errorf("aborting after updating NextRound=%d and failing to save metadata: %w", lastRnd+1, lastError)
+					return
+				}
+				p.logger.Tracef("exporter %s @ round %d saved pipeline metadata", exporter.Metadata().Name, p.pipelineMetadata.NextRound)
+
+				for i, cb := range p.completeCallback {
+					p.logger.Tracef("exporter %s @ round=%d NextRound=%d executing callback %d", exporter.Metadata().Name, lastRnd, p.pipelineMetadata.NextRound, i)
+					_, lastError = RetriesNoOutput(cb, blkData, p, fmt.Sprintf("callback %d", i))
```
tzaffi, line 599: Should we send metrics for the callbacks?

conduit/pipeline/pipeline.go
```diff
@@ -431,8 +444,216 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
+func (p *pipelineImpl) ExporterHandler(exporter exporters.Exporter, blkChan <-chan data.BlockData) {
+	p.wg.Add(1)
+	go func() {
+		defer p.wg.Done()
+		var lastError error
+		lastRnd := uint64(0)
+		totalSelectWait := time.Duration(0)
+		totalExportWork := time.Duration(0)
+
+		defer func() {
```
tzaffi, line 550: TODO: merge these defers

conduit/pipeline/pipeline.go
```diff
@@ -431,8 +455,220 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
+			}
+		}
+	}(p.pipelineMetadata.NextRound)
+
+	<-p.ctx.Done()
+	// TODO: send a prometheus observation based on WhyStopped()
+}
+
+// Start pushes block data through the pipeline
+func (p *pipelineImpl) OStart() {
```
tzaffi, line 671: TODO: remove this function and any orphaned code before merging!!!!

conduit/pipeline/pipeline.go
```diff
@@ -59,16 +61,28 @@ type pipelineImpl struct {
...
-	p.mu.Lock()
-	defer p.mu.Unlock()
-	p.err = err
+func (p *pipelineImpl) joinError(err error) {
+	p.errMu.Lock()
+	defer p.errMu.Unlock()
+	p.err = errors.Join(p.err, err)
+}
+
+func (p *pipelineImpl) cancelWithProblem(err error) {
```
tzaffi, line 83: For example, here's a final conduit error resulting from an importer problem:

```
Exiting with error:     importer algod handler (importer cancelled): failed to import round 103 after 11003ms: algod: giving up after 1 retries: wrong round returned from status for round: 102 != 103: status2.LastRound mismatch: context deadline exceeded
```

pkg/cli/cli.go
```diff
@@ -134,7 +134,7 @@ Detailed documentation is online: https://github.com/algorand/conduit`,
 		Run: func(cmd *cobra.Command, args []string) {
 			err := runConduitCmdWithConfig(cfg)
 			if err != nil {
-				fmt.Fprintf(os.Stderr, "\nExiting with error:\n%s.\n", err)
+				fmt.Fprintf(os.Stderr, "\nExiting with error:\t%s.\n", err)
```
tzaffi, line 137: This improves the clarity of the E2E tests in the case of an error.

conduit/data/config.go
```diff
@@ -10,6 +10,17 @@ import (
 	yaml "gopkg.in/yaml.v3"
 )
 
+const (
```
tzaffi, line 13: TODO: pull in #119 instead of these changes once that is merged

conduit/pipeline/pipeline.go
```diff
@@ -431,8 +455,225 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
 	metrics.ImportedTxnsPerBlock.Observe(float64(len(block.Payset)) + float64(innerTxn))
 }
 
+func (p *pipelineImpl) ImportHandler(importer importers.Importer, roundChan <-chan uint64, blkOutChan chan<- data.BlockData) {
```
tzaffi, line 458: TODO: shouldn't the handlers be private?

e2e_tests/src/e2e_conduit/runner.py
```diff
@@ -47,6 +47,7 @@ def setup_scenario(self, scenario):
             yaml.dump(
                 {
                     "log-level": "info",
+                    "retry-count": 1,
```
tzaffi, line 50: Formerly this was using the default value of 10

e2e_tests/src/e2e_conduit/subslurp.py
```diff
@@ -6,29 +6,28 @@
...
         self.f = f
         self.buf = io.BytesIO()
         self.gz = gzip.open(self.buf, "wb")
         self.timeout = timedelta(seconds=120)
-        # Matches conduit log output: "Pipeline round: 110"
-        self.round_re = re.compile(b'.*"Pipeline round: ([0-9]+)"')
         self.round = 0
         self.error_log = None
 
-    def logIsError(self, log_line):
```
tzaffi, line 22: Inlining this check in `tryParseRound()`

conduit/pipeline/pipeline.go
tzaffi: I'm keeping it around while I'm continuing more local testing.

conduit/pipeline/pipeline.go
```diff
@@ -431,8 +455,228 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
+						lastError = fmt.Errorf("aborting due to failed callback %d: %w", i, lastError)
+						return
+					}
+				}
+				lastError = nil
+				// WARNING: removing the following will BREAK:
+				// - the E2E test (Search for "Pipeline round:" in subslurp.py)
+				// - the internal tools logstats collector (See func ConduitCollector in logstats.go of internal-tools repo)
+				// Modify with CAUTION!!!!
+				p.logger.Infof("UPDATED Pipeline NextRound=%d. FINISHED Pipeline round r=%d (%d txn) exported in %s", nextRound, lastRound, len(blkData.Payset), time.Since(startStart))
```
tzaffi, line 626: Here's a [playground](https://go.dev/play/p/_idxxpGBlYL) showing that this new log message should still work with the `logstats.go` regexp.

shiqizng: is the time.Since(startStart) correct after the 1st block? `startStart` is the time when the pipeline started and it never gets updated, right?

conduit/pipeline/pipeline.go
```diff
@@ -372,7 +386,7 @@ func (p *pipelineImpl) Init() error {
 }
 
 func (p *pipelineImpl) Stop() {
-	p.cf()
+	p.ccf(errStopCause)
```
shiqizng, line 389: nit: rename the var? the message `"pipeline stopped"` doesn't seem like an error?

conduit/pipeline/pipeline.go
tzaffi: This is a good point. In fact, the [original code](https://github.com/algorand/conduit/blob/master/conduit/pipeline/pipeline.go#L474-L498) was logging the time since the importer was completed, not the time from the beginning of `Start()`. Unfortunately, now that channels intermediate between plugins we can't easily access this value as we did in the original code. If this statistic is important enough, instead of having the channels be of type `chan data.BlockData` they could be of type `chan struct{importFinishTime time.Time, blk data.BlockData}`.  What do you think?

conduit/pipeline/pipeline.go
```diff
@@ -59,16 +61,28 @@ type pipelineImpl struct {
 	pipelineMetadata state
 }
 
+var (
```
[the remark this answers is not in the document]
tzaffi: When I change `var` to `const` I get an error: `errors.New("pipeline stopped") (value of type error) is not constant` and that makes sense because constants need to be defined at compile-time. On the other hand, wrapping them inside of functions won't work either, since I want to be able to use `errors.Is()` with them.

conduit/pipeline/common.go
```diff
@@ -1,10 +1,87 @@
...
 func HandlePanic(logger *log.Logger) {
 	if r := recover(); r != nil {
 		logger.Panicf("conduit pipeline experienced a panic: %v", r)
 	}
 }
+
+type empty struct{}
+
+type pluginInput interface {
+	uint64 | data.BlockData | string | empty
```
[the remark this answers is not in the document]
tzaffi: The reason is that I'm using `pluginInput` for both parameters and result values. Result values can be `empty`. I think part of the confusion is my usage of the name `pluginInput`. I could see 2 possible improvement:
* rename `pluginInput` to `pluginParamOrResult`
* introduce a separate type `pluginOutput` which has `empty` and remove `empty` from `pluginInput`

Which one do you prefer?

conduit/pipeline/common.go
```diff
@@ -1,10 +1,87 @@
...
+		}
+		opStart := time.Now()
+		y2, err2 := f(x)
+		if err2 == nil {
+			return y2, time.Since(opStart), nil
+		}
+
+		p.logger.Infof("%s: retry number %d/%d with err: %v", msg, i, p.cfg.RetryCount, err2)
+		if p.cfg.RetryCount > 0 {
+			// TODO: this feels like a code smell. Probly better to always keep only the last error.
```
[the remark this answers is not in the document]
tzaffi: will get rid of the Join.

conduit/pipeline/pipeline.go
shiqizng: or we can add a timestamp data.BlockData?

conduit/pipeline/common_test.go
```diff
@@ -0,0 +1,210 @@
...
+		{
+			name:         "retry 5 succeeds after 5",
+			retryCount:   5,
+			succeedAfter: 5,
+			neverSucceed: false,
+		},
+		{
+			name:         "retry 5 succeeds after 7",
+			retryCount:   5,
+			succeedAfter: 7,
```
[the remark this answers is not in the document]
tzaffi: This is a bit confusing. `retryCount` belongs to the pipeline. `succeedAfter` belongs to the factory produced function that will get retried. `succeedAfter = 7`, it means it will succeed on 8th attempt (or 7th retry). For this test case, since `retryCount = 5` we expect that `Retries(theFunction, ...)` should fail because it gives up before the 7th retry.

conduit/pipeline/common_test.go
```diff
@@ -0,0 +1,210 @@
...
+				require.ErrorIs(t, err, errTestCancelled, tc.name)
+				require.ErrorIs(t, err, errSentinelCause, tc.name)
+				require.Zero(t, y, tc.name)
+				return
+			}
+
+			y, _, err := Retries(succeedAfter, 0, p, "test")
+			if tc.retryCount == 0 { // WLOG tc.neverSucceed == false
+				require.NoError(t, err, tc.name)
+				require.Equal(t, tc.succeedAfter, y-1, tc.name)
```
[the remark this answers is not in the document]
tzaffi: The problem is that for the test I want `y` to only be `uint64`'s zero-value when `Retries()` fails (see the comment on line 24^^^). So in order to do that I add +1 which then needs to be subtracted out here.

However, I'll add a comment here explaining the convoluted arithmetic.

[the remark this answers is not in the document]
tzaffi: (:eyes are to remind me that I'm promising something)

conduit/pipeline/pipeline.go
tzaffi: I totally agree that this isn't an error situation. However, cancellation causes must be errors and the linter doesn't like me creating errors that I don't name with `err` prefix. Shall I tray to figure out how to get the linter to stop complaining?

conduit/pipeline/pipeline.go
```diff
@@ -431,8 +455,225 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
 
+func (p *pipelineImpl) ImportHandler(importer importers.Importer, roundChan <-chan uint64, blkOutChan chan<- data.BlockData) {
+	p.wg.Add(1)
+	go func() {
+		defer p.wg.Done()
+		lastRnd := uint64(0)
+		totalSelectWait := time.Duration(0)
+		totalFeedWait := time.Duration(0)
+		defer func() {
+			p.logger.Debugf("importer handler exiting. lastRnd=%d totalSelectWait=%dms; totalFeedWait=%dms", lastRnd, totalSelectWait.Milliseconds(), totalFeedWait.Milliseconds())
```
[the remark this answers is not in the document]
tzaffi: I believe it suffers the same fate as [this example](https://go.dev/play/p/MjOfa2bDxiF)

conduit/pipeline/pipeline.go
tzaffi: Maybe embed new `type PipelineData struct ...` inside of `BlockData` ?

conduit/pipeline/pipeline_test.go
```diff
@@ -249,59 +267,84 @@ func mockPipeline(t *testing.T, dataDir string) (*pipelineImpl, *test.Hook, *moc
...
 			NextRound:   0,
 			GenesisHash: "",
 		},
 		cfg: &data.Config{
 			RetryDelay: 0 * time.Second,
 			RetryCount: math.MaxUint64,
+			Processors: []data.NameConfigPair{{
+				Name:   "mockProcessor",
+				Config: map[string]interface{}{},
+			}},
```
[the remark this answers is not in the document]
tzaffi: This was getting an index-out-of-bounds panic because I introduced `Init()` on line 320 below. Then inside of `Init()` there's [an expectation](https://github.com/tzaffi/conduit/blob/4f5576352b6cad9eee190883f5af0b4dbe1e2cc4/conduit/pipeline/pipeline.go#L351) that the pipeline's  config has a slice storing the processor configs.

e2e_tests/src/e2e_conduit/subslurp.py
```diff
@@ -6,29 +6,28 @@
 
 logger = logging.getLogger(__name__)
 
+# Matches conduit log output:
+# "FINISHED Pipeline round: 110. UPDATED Pipeline round: 111"
+FINISH_ROUND: re.Pattern = re.compile(
```
tzaffi, line 11: TODO: something's buggy here as the E2E test is failing.

conduit/pipeline/common.go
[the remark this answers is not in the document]
tzaffi: For now I went with the second approach (introduce `pluginOutput` type): 2c2e38702b4b647b0bcac64ae112126cd0f23202

conduit/pipeline/common.go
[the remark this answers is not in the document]
tzaffi: 2c2e38702b4b647b0bcac64ae112126cd0f23202

conduit/pipeline/pipeline_test.go
```diff
@@ -249,59 +267,84 @@ func mockPipeline(t *testing.T, dataDir string) (*pipelineImpl, *test.Hook, *moc
...
 // TestPipelineRun tests that running the pipeline calls the correct functions with mocking
 func TestPipelineRun(t *testing.T) {
-	mImporter := mockImporter{}
-	mImporter.On("GetBlock", mock.Anything).Return(uniqueBlockData, nil)
+	mImporter := mockImporter{rndOverride: initialRound + 1}
+	mImporter.On("GetBlock", mock.Anything).Return(mock.Anything, nil)
+	mImporter.On("OnComplete", mock.Anything).Return(nil)
 	mProcessor := mockProcessor{}
 	processorData := uniqueBlockData
 	processorData.BlockHeader.Round++
```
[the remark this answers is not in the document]
tzaffi: d2e0d3da5b7403c830d87fca3cf5307cb206cedb

conduit/pipeline/common_test.go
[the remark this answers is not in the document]
tzaffi: b2e67de6c5147b57a37b5e9334e6214489ba90ab

conduit/pipeline/pipeline.go
```diff
@@ -431,8 +455,225 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
+			selectStart := time.Now()
+			select {
+			case <-p.ctx.Done():
+				p.logger.Infof("importer handler exiting. lastRnd=%d", lastRnd)
+				return
+			case rnd := <-roundChan:
+				lastRnd = rnd
+				waitTime := time.Since(selectStart)
+				totalSelectWait += waitTime
+				p.logger.Tracef("importer handler waited %s to receive round %d", waitTime, rnd)
```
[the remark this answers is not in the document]
tzaffi: resolving as I believe this is out of date

conduit/pipeline/pipeline.go
tzaffi: VS Code still complains, but it should be passing lint now: deec57e61df93b3293e03eae9871ae8464785418

conduit/pipeline/pipeline.go
tzaffi: 4103de1c50b8ef0c89af9ec3197437b9808b067f

tzaffi: I had a change of mind. I don't want to embed `PipelineData` inside of `data.BlockData`. That is mixing together 2 totally different things and even has a risk of "infecting" the indexer database (e.g. suppose the queries naively serialize the entirety of `data.BlockData` into json and just save as is...). So I'll be moving `PipelineData` into the `pipeline` package and creating something like `type block struct` which itself embeds a `data.BlockData`.

conduit/data/block_export_data.go
```diff
@@ -21,9 +21,10 @@ type InitProvider interface {
 	GetTelemetryClient() telemetry.Client
 }
 
+// PipelineData is used to keep track of pipeline performance
+
```
tzaffi, line 25:
```suggestion
```

conduit/pipeline/pipeline.go
tzaffi: 5ae8a21baaabd327d143c18f68efa79338ef3338

conduit/pipeline/pipeline.go
tzaffi: 5ae8a21baaabd327d143c18f68efa79338ef3338

conduit/pipeline/pipeline.go
```diff
@@ -431,8 +467,250 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
+					len(pluginBlk.Payset),
+					time.Since(pluginBlk.finishImportTime),
+				)
+			}
+		}
+	}()
+}
+
 // Start pushes block data through the pipeline
 func (p *pipelineImpl) Start() {
```
winder, line 668: If you pass a context in here could we get rid of the `Stop` and `WhyStopped` functions? If yes, that might be a good followup PR.

winder (commented):
Some initial feedback.

conduit/pipeline/pipeline.go
```diff
@@ -372,7 +398,7 @@ func (p *pipelineImpl) Init() error {
 }
 
 func (p *pipelineImpl) Stop() {
-	p.cf()
+	p.ccf(stopCause)
```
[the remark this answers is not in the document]
tzaffi: This is a good point. If the only way cancel under non-error conditions is through `Stop()` or we don't need to disambiguate, we should use `nil` here. On the other hand, if we have other ways to _happy-path-cancel_ the context, and want to tell them apart, then we'd need this approach. I don't have a strong opinion here.

conduit/pipeline/pipeline.go
```diff
@@ -405,6 +431,16 @@ func (p *pipelineImpl) Stop() {
 	}
 }
 
+// WhyStopped returns nil if a context was never provided or the pipeline
+// is yet to have stopped. Otherwise, it returns the cause of the pipeline's
+// context cancellation.
+func (p *pipelineImpl) WhyStopped() error {
```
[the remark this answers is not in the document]
tzaffi: Yes. I was motivated to elevate this capability in the API, but I'll remove it as it's not actually necessary once you know about cancellation causes.

conduit/pipeline/pipeline.go
```diff
@@ -431,8 +467,250 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
+					// Previously we reported time starting after block fetching is complete
+					// through the end of the export operation. Now that each plugin is running
+					// in its own goroutine, report only the time of the export.
+					addMetrics(pluginBlk.BlockData, exportTime)
+				}
+
+				// Increment Round, update metadata
+				nextRound := lastRound + 1
+				p.pipelineMetadata.NextRound = nextRound
+				_, lastError = RetriesNoOutput(p.pipelineMetadata.encodeToFile, p.cfg.ConduitArgs.ConduitDataDir, p, "pipelineMetadata save")
```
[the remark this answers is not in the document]
tzaffi: I'll fail after the first error.

conduit/pipeline/pipeline.go
```diff
@@ -431,8 +467,250 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
+				metrics.ExporterTimeSeconds.Observe(exportTime.Seconds())
+				// Ignore round 0 (which is empty).
+				if lastRound > 0 {
+					// Previously we reported time starting after block fetching is complete
+					// through the end of the export operation. Now that each plugin is running
+					// in its own goroutine, report only the time of the export.
+					addMetrics(pluginBlk.BlockData, exportTime)
+				}
+
+				// Increment Round, update metadata
```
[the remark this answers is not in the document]
tzaffi: Since this is happening inside of a go-routine, to keep everything consistent (NextRound should be updated and metadata save before the next exporter is completed) I would need to coordinate with a channel. Do you think this is worthwhile?

conduit/pipeline/pipeline.go
```diff
@@ -431,8 +467,250 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
+				// - the internal tools logstats collector (See func ConduitCollector in logstats.go of internal-tools repo)
+				// Modify with CAUTION!!!!
+				p.logger.Infof(
+					"UPDATED Pipeline NextRound=%d after [%s] from round kickoff. FINISHED Pipeline round r=%d (%d txn) exported in %s",
+					nextRound,
+					time.Since(pluginBlk.startRoundTime),
+					lastRound,
+					len(pluginBlk.Payset),
+					time.Since(pluginBlk.finishImportTime),
+				)
```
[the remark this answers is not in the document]
tzaffi: If we take the end of round work out of the exporter, then we'll need a return channel -so that dovetails with your comment. But if we want to keep a semantically equivalent log message (for use by internal tools `logstats.go`) the timer needs to start at the end of the import, and needs to be sent down through the channels so I don't see how to avoid something like this.

In my current EC2 testing, I haven't found the extra log information very useful. In fact, it can be misleading because there is work happening concurrently so you can't just add up the processing times and figure out the contribution to the overall time.

My proposal is:
1. revert to channels of type `chan data.BlockData`
2. exporter provides a return channel of this type
3. roundFinishHandler reads the channel. In particular it:
   * does each portion of the work _failing fast without any retries_
   * logs the "...FINISHED Pipeline round r=..." message
4. modify `logstats.go` so that it _doesn't care_ about the time to export
5. we can add more TRACE logs to get precise metrics if we need in the future

conduit/pipeline/pipeline.go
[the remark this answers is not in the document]
winder: I'll have to think about this more.

There is a race condition between the exporter finishing and metadata.json being written where the service could be terminated. So the exporter could be called multiple times for the same round. I guess what I'm suggesting could allow the exporter to be called for the previous round. That doesn't sound good. An unbuffered channel could synchronize it a little bit by blocking the exporter from writing results until the callbacks execute, but it would be one round ahead of the callbacks.

conduit/pipeline/pipeline.go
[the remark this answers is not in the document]
tzaffi: There is a more ambitious approach which is to have a metrics collector keep track of all the times each plugin starts/stops and then do its own calculations accordingly. But that seems like too much of a rabbit hole to go into at this juncture.

conduit/pipeline/pipeline.go
tzaffi: `WhyStopped()` I'm removing regardless. And I could see a followup to remove `Stop()` as you suggest.

winder: `Stop` seems fine to keep for the interface, since it would let things like tests skip using the context

conduit/pipeline/pipeline.go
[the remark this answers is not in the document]
tzaffi: Yes, callbacks as well are part of the `finishRound()` logic.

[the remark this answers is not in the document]
winder: I think you've convinced me that this should stay here. The timing metadata collected from other plugins still seems out of place, I think you can resolve this thread since we've got another about that already.

conduit/pipeline/pipeline.go
```diff
@@ -431,8 +467,250 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
+			p.logger.Tracef("pushing round %d into roundChan", rnd)
+			select {
+			case <-p.ctx.Done():
+				p.logger.Infof("round channel feed exiting. lastRnd=%d", rnd)
+				return
+			case roundChan <- rnd:
+				waitTime := time.Since(selectStart)
+				totalFeedWait += waitTime
+				rnd++
+			}
```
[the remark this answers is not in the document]
tzaffi: I think this logic would need to go into the pipeline and having it be done in a select here would be too non-deterministic.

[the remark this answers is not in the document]
winder: agreed. Resolving.

conduit/pipeline/pipeline.go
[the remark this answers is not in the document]
shiqizng: I see.

conduit/pipeline/pipeline.go
```diff
@@ -431,8 +467,250 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
+					lastError = fmt.Errorf("aborting after failing to export round %d: %w", lastRound, lastError)
+					return
+				}
+				metrics.ExporterTimeSeconds.Observe(exportTime.Seconds())
+				// Ignore round 0 (which is empty).
+				if lastRound > 0 {
+					// Previously we reported time starting after block fetching is complete
+					// through the end of the export operation. Now that each plugin is running
+					// in its own goroutine, report only the time of the export.
+					addMetrics(pluginBlk.BlockData, exportTime)
```
shiqizng, line 628: why not add an import timestamp to BlockData? then you could have 
`addMetrics(blockdata, time.Since(blockdata.importTime))`

tzaffi: After our previous discussion, I temporarily added **importTime** to `blockdata` . Then I decided I didn't like polluting block chain crucial data with performance metrics and also imposing a requirement on downstream processes to know about this pollution (imagine an exporter that just serializes all of the `BlockData` and then gets this extra info it wasn't expecting). There's a lively discussion related to this [on a thread above](https://github.com/algorand/conduit/pull/128#discussion_r1284363465).

conduit/pipeline/pipeline.go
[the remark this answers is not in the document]
Eric-Warehime: I feel like you've looked into graceful pipeline exits before/whether we properly respond to things like sigterm by running some shutdown logic. In that case I would expect to attempt to call Stop and for it to return without error right?

conduit/pipeline/pipeline.go
```diff
@@ -431,8 +467,250 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
 	metrics.ImportedTxnsPerBlock.Observe(float64(len(block.Payset)) + float64(innerTxn))
 }
 
+func (p *pipelineImpl) importerHandler(importer importers.Importer, roundChan <-chan uint64, blkOutChan pluginChannel) {
+	p.wg.Add(1)
+	go func() {
```
Eric-Warehime, line 472: Why did you decide to launch the goroutines inside of each of the plugin functions instead of writing them as regular functions and having `Start` launch all of them as goroutines?

conduit/pipeline/common_test.go
```diff
@@ -0,0 +1,213 @@
...
+		return func(x uint64) error {
+			if tries >= succeedAfter && !never {
+				return nil
+			}
+			tries++
+			return fmt.Errorf("%w: tries=%d", errSentinelCause, tries-1)
+		}
+	}
+
+	cases := []struct {
```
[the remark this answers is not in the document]
tzaffi: I'm a general fan of property based tests / fuzzing in general. It could be useful here as well because I may have missed an edge case. But I do have confidence that the test cases provide reasonable coverage and gained further confidence after observing the catchup on EC2.

conduit/pipeline/pipeline.go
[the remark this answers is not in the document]
tzaffi: It's a good juncture to experiment a bit and see how to get the graceful exit to work properly, and this will hopefully lead the way to the simplest cancellation approach. Some references:
* #100 
* #97 
*

conduit/pipeline/pipeline.go
tzaffi: I can see though, that there could be a convention by which all go-routines launched from a function ought to be launched inside that man function with `go xyz()`... If we want to follow such a convention, I can refactor.

conduit/pipeline/pipeline.go
tzaffi: going going...

tzaffi: gone

conduit/pipeline/pipeline.go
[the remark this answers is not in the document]
tzaffi: Upon further thought, I've changed my mind having remembered the original purpose. Consider the [conduit client](https://github.com/tzaffi/conduit/blob/7e88db3d2c9fea7afc600b0cee46f1f7c4e4fffd/pkg/cli/cli.go#L103). It may want to know why the pipeline stopped but it only has access to the `Pipeline interface`. In fact, instead of the current `defer pipeline.Stop()` call, we may want to have:

```go
def func() {
    pline.Stop()
    if !errors.Is(pline.WhyStopped(), pipeline.BecauseStopMethod) {
        logger.Error("unexpected conduit pipeline exit: %v", pline.WhyStopped())
    }
}()
```

conduit/pipeline/pipeline.go
```diff
@@ -431,103 +469,252 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
+				p.logger.Infof("importer handler exiting. lastRnd=%d", lastRnd)
 				return
+			case rnd := <-roundChan:
+				lastRnd = rnd
+				waitTime := time.Since(startRound)
+				totalSelectWait += waitTime
+				p.logger.Tracef("importer handler waited %dms to receive round %d", waitTime.Milliseconds(), rnd)
+
+				// TODO: DO NOT MERGE THE FOLLOWING!!!!
+				// time.Sleep(5 * time.Second)
```
tzaffi, line 496:
```suggestion
```

conduit/pipeline/pipeline.go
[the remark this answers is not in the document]
tzaffi: I've followed this suggestion. In particular:
* fail fast if can't persist the metadata/next round
* carry on if the callbacks fail - this is motivated by the vanilla use case where the only callback is algod's `/sync` request, which as 300+ rounds to succeed.

I've also modified some docs - @algoanne 

cb86036bd9a7825c95d594e3d610f7a4171c9d16

conduit/plugins/importers/algod/algod_importer.go
```diff
@@ -42,7 +42,7 @@ const (
 )
 
 var (
-	waitForRoundTimeout = 5 * time.Second
+	waitForRoundTimeout = 15 * time.Second
```
tzaffi, line 45: TODO: import #133 which trumps this change.

docs/PluginDevelopment.md
```diff
@@ -62,11 +62,15 @@ The context provided to this function should be saved, and used to terminate any
 
 ### Per-round function
 
-Each plugin type has a function which is called once per round:
+Each plugin type has a function termed the _Plugin Method_ which is called once per round:
```
tzaffi, line 65: @algoanne WDYT?

conduit/pipeline/pipeline.go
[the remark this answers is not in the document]
tzaffi: Upon further reflection, I'm liking a more streamlined interface as suggested in [this comment](https://github.com/algorand/conduit/pull/97/files#r1227096289).

So we get  rid of `Stop()`, but keep `WhyStopped()`, and also have the client be in charge of listening to interrupt signals and and issuing cancellations it nows about and can ignore using `WhyStopped()`.

conduit/pipeline/pipeline.go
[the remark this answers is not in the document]
tzaffi: Let's continue this discussion on [this thread](https://github.com/algorand/conduit/pull/128#discussion_r1289543345)

tzaffi:
> If it's easy for you to run, it might be interesting to see `Local test network 5 minute sprint` using different log levels to show how logrus/logging in general is impacting processing times.

Can we add this as a task for #131 ?

conduit/pipeline/pipeline.go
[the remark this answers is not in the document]
tzaffi: I didn't get ride of `Stop()`, but did get rid of `WhyStopped()`. I added more questions probing the possible resurrection of these to #100 

see: 1158f757ad8429bca46dc2ce96d7eee4fa77f639

conduit/pipeline/pipeline.go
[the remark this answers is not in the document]
tzaffi: I'm proposing only recording the "exporter". Unfortunately, when the importer < exporter, under pipelining we can no longer deduce the "importer" time. However, a future possibility is to for the importer to add a separate log.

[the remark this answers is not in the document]
tzaffi: 867973f82e586f7d736e53e44eaf6d1f08871275

conduit/pipeline/common.go
```diff
@@ -1,10 +1,83 @@
...
+// - Upon success:
+//   - a nil error is returned even if there were intermediate failures
+//   - the returned duration dur measures the time spent in the call to f() that succeeded
+//
+// - Upon failure:
+//   - the return value y is the zero value of type Y and a non-nil error is returned
+//   - when p.cfg.retryCount > 0, the error will be a join of all the errors encountered during the retries
+//   - when p.cfg.retryCount == 0, the error will be the last error encountered
+//   - the returned duration dur is the total time spent in the function, including retries
+func Retries[X pluginInput, Y pluginOutput](f func(x X) (Y, error), x X, p *pipelineImpl, msg string) (y Y, dur time.Duration, err error) {
```
winder, line 48: Consider:
* breaking out the context.Context out into a parameter, I was specifically looking for it to make sure the wait is properly terminated.
* also breaking out RetryCount and logger. It's already a pure function in how it's used, but the signature doesn't guarantee that it's pure.

conduit/pipeline/pipeline.go
```diff
@@ -431,103 +445,243 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
+				importFinish := time.Now()
+
+				select {
+				case <-p.ctx.Done():
+					return
+				case blkOutChan <- blkData:
+					waitTime := time.Since(importFinish)
+					totalFeedWait += waitTime
+					p.logger.Tracef("imported round %d into blkOutChan after waiting %dms on channel", rnd, waitTime.Milliseconds())
+				}
```
winder, line 490: Also in the other handlers, this could be helpful
```suggestion
				// check context in case out channel is full.
				select {
				case <-p.ctx.Done():
					return
				case blkOutChan <- blkData:
					waitTime := time.Since(importFinish)
					totalFeedWait += waitTime
					p.logger.Tracef("imported round %d into blkOutChan after waiting %dms on channel", rnd, waitTime.Milliseconds())
				}
```

conduit/pipeline/pipeline.go
```diff
@@ -431,103 +445,243 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
+					return
+				}
+				metrics.ImporterTimeSeconds.Observe(importTime.Seconds())
+				p.logger.Tracef("importer handler imported round %d in %dms", rnd, importTime.Milliseconds())
+
+				// TODO: Verify that the block was built with a known protocol version.
+
+				importFinish := time.Now()
+
+				select {
```
winder, line 483:
```suggestion
				// check context in case out channel is full.
				select {
```

conduit/pipeline/pipeline.go
```diff
@@ -431,103 +445,243 @@ func addMetrics(block data.BlockData, importTime time.Duration) {
...
+				var lastError error
+				blk, procTime, lastError = Retries(proc.Process, blk, p, proc.Metadata().Name)
+				if lastError != nil {
+					p.cancelWithProblem(fmt.Errorf("processor[%d] %s handler (%w): failed to process round %d after %dms: %w", idx, proc.Metadata().Name, errProcessorCause, lastRnd, procTime.Milliseconds(), lastError))
+					return
+				}
+				metrics.ProcessorTimeSeconds.WithLabelValues(proc.Metadata().Name).Observe(procTime.Seconds())
+
+				selectStart := time.Now()
+				select {
```
winder, line 530:
```suggestion
				// check context in case out channel is full.
				select {
```

pkg/cli/cli.go
```diff
@@ -80,7 +80,7 @@ func runConduitCmdWithConfig(args *data.Args) error {
 	}
 
 	ctx := context.Background()
-	pipeline, err := pipeline.MakePipeline(ctx, pCfg, logger)
```
tzaffi, line 83: No functional changes here, but removing the shadowing of the package by the variable.

shiqizng (approved):

PR #139, opened 2023-08-16
api: New API with health endpoint

winder:
## Summary

Add a new option for an API which includes a health endpoint to return the current round.

[Demo and overview.](https://www.youtube.com/watch?v=wDRf_haonUk)

## Test Plan

New unit tests.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -49,7 +49,7 @@ func StartServer(logger *log.Logger, p StatusProvider, address string) (func(ctx
 	}()
 
 	shutdownCallback := func(ctx context.Context) {
-		if err := srv.Shutdown(ctx); err != nil {
+		if err := srv.Shutdown(ctx); err != nil && !errors.Is(err, http.ErrServerClosed) {
 			logger.Fatalf("failed to shutdown API server: %s", err)
 		}
 	}
--- a/api/api.go
+++ b/api/api.go
@@ -31,7 +31,7 @@ func makeHealthHandler(p StatusProvider) func(w http.ResponseWriter, r *http.Req
 	}
 }
 
-// StartServer starts a http server that exposes a health check endpoint.
+// StartServer starts an http server that exposes a health check endpoint.
 // A callback is returned that can be used to gracefully shutdown the server.
 func StartServer(logger *log.Logger, p StatusProvider, address string) (func(ctx context.Context), error) {
 	mux := http.NewServeMux()
--- a/api/api.go
+++ b/api/api.go
@@ -1,0 +1,57 @@
+package api
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/algorand/conduit/conduit/pipeline"
+)
+
+// StatusProvider is a subset of the Pipeline interface required by the health handler.
+type StatusProvider interface {
+	Status() (pipeline.Status, error)
+}
+
+func makeHealthHandler(p StatusProvider) func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		status, err := p.Status()
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			fmt.Fprintf(w, `{"error": "%s"}`, err)
+			return
+		}
+		w.WriteHeader(http.StatusOK)
+		data, _ := json.Marshal(status)
+		fmt.Fprint(w, string(data))
+	}
+}
+
+// StartServer starts a http server that exposes a health check endpoint.
+// A callback is returned that can be used to gracefully shutdown the server.
+func StartServer(logger *log.Logger, p StatusProvider, address string) (func(ctx context.Context), error) {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/health", makeHealthHandler(p))
+
+	srv := &http.Server{
+		Addr:    address,
+		Handler: mux,
+	}
+
+	go func() {
+		if err := srv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			logger.Fatalf("failed to start API server: %s", err)
+		}
+	}()
+
+	shutdownCallback := func(ctx context.Context) {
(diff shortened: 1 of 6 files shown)


conduit/pipeline/pipeline.go
```diff
@@ -532,6 +539,13 @@ func (p *pipelineImpl) Wait() {
 	p.wg.Wait()
 }
 
+func (p *pipelineImpl) Status() (Status, error) {
+	rnd := atomic.LoadUint64(&p.pipelineMetadata.NextRound)
```
algochoi, line 543: Docs seem to suggest using `Load()` here: https://pkg.go.dev/sync/atomic#LoadUint64

Also do we need to atomically increment it here? https://github.com/winder/conduit/blob/3c92877960bbc00f26070a3076a653e3c5da46f7/conduit/pipeline/pipeline.go#L462

algochoi (commented):
I like the video demo, very neat and we should do it more often.
I also tried it out on sandbox, but failed -- it looks like the metrics endpoint is exposed, is there a similar thing we can do for `/health` or am I just misusing it?

conduit/pipeline/pipeline.go
winder: Zeph is working on that code, so I didn't want to touch it here. We're careful to avoid using `pipelineMetadata` in other go routines, so I don't think we need any special handling there.

I saw that comment in the docs, but wasn't sure I wanted to change types for something so simple. Looking again [and now finding the bugs reference](https://pkg.go.dev/sync/atomic#pkg-note-BUG), maybe that would be a better option.

winder: @tzaffi what do you think about this?

tzaffi: I think you should be able to use `p.pipeline.NextRound.Load()` here without interfering so much with #128. However, I do agree that refactoring all the usages is worthwhile and I created an issue: #140 .

tzaffi (commented):
still working through the review, but I was pinged and don't want to keep folks waiting

api/api.go
```diff
@@ -0,0 +1,57 @@
...
+
+	go func() {
+		if err := srv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			logger.Fatalf("failed to start API server: %s", err)
+		}
+	}()
+
+	shutdownCallback := func(ctx context.Context) {
+		if err := srv.Shutdown(ctx); err != nil {
+			logger.Fatalf("failed to shutdown API server: %s", err)
```
tzaffi, line 53: I'm a bit concerned about panicking here if there is an error. For example (see https://pkg.go.dev/net/http#Server.Shutdown) if the server is shutdown twice, the 2nd time would issue a `ErrServerClosed`. Maybe it's worth to condition the panic on not being this error.

conduit/data/config.go
```diff
@@ -56,6 +56,11 @@ type Telemetry struct {
 	Password string `yaml:"password"`
 }
 
+// API defines parameters for the Conduit API server.
+type API struct {
+	Address string `yaml:"addr"`
```
tzaffi, line 61: If all we really need or can handle is the port, maybe just change `Address` &rarr; `Port` ?

api/api_test.go
```diff
@@ -0,0 +1,103 @@
...
+	require.Len(t, h.Entries, 1)
+	require.Equal(t, h.LastEntry().Level, logrus.FatalLevel)
+}
+
+func TestStartServer_GracefulShutdown(t *testing.T) {
+	l, h := test.NewNullLogger()
+	// capture the fatal log if any.
+	l.ExitFunc = func(int) {}
+	sp := mockStatusProvider{}
+	shutdown, err := StartServer(l, sp, "bad address")
```
tzaffi, line 52: Per other discussion, add a 2nd shutdown attempt to verify no panic.

conduit/pipeline/pipeline.go
algochoi: I don't feel very strongly since getting `round` wrong is probably not critical to most applications. I'll defer judgement to you.

conduit/pipeline/pipeline.go
```diff
@@ -64,6 +63,9 @@ type pipelineImpl struct {
 	completeCallback []conduit.OnCompleteFunc
 
 	pipelineMetadata state
+
+	statusMu sync.Mutex
```
tzaffi, line 67: With this change, can we wait until after #128 goes in before merging this PR? It may require an additional EC2 test.

conduit/pipeline/pipeline.go
winder: @algochoi @tzaffi The comments about endienness and caveats about what processors support 64 bit atomic operations seemed dangerous, so I went ahead and switched to a simple mutex.

A channel could be used to keep the pipeline loop lock-free, but given the lock only needs to be held for the duration of a copy or an assignment, it didn't seem like that would have any real impact to performance.

conduit/data/config.go
winder: I had it that way originally, but the metrics were defined as a string, and so are the addresses in Indexer/Algod. As a string you can do things like use ipv6 addresses.

api/api_test.go
```diff
@@ -0,0 +1,103 @@
...
+	defer shutdown(context.Background())
```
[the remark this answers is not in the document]
winder: just making sure the server is shutdown after the test completes

api/api.go
winder: Good point. Because this is returned as a callback the user might call it multiple times. I added tthe `ErrServerClosed` check to avoid that problem.

api/api_test.go
winder: It was definitely exiting / calling the `Fatalf` log. Thats why `ExitFunc` was needed for this test.

conduit/pipeline/pipeline.go
winder: Sure. Adding a mutex was much less disruptive than switching to `atomic.Uint64`.

tzaffi (commented):
IMO, this is approvable, but let's wait for #128 to get merged in first.

algochoi (approved):
LGTM -- do we need to do something special to expose this endpoint in sandbox environments?

PR #142, opened 2023-08-18
file-plugins: enabling messagepack format

tzaffi:
## File Plugins: Add Messagepack Format and Default to `msgp.gz`

Based on the extension of the file pattern provided in a `file_reader` or `file_writer` plugin, choose:
1. the **Encoding Format** (based on `.msgp` or `.json`)
2. whether compression is used (based on `.gz`)

When no pattern is provided, it defaults to `*.msgp.gz`

## Test Plan

This is all tested in CI including a new integration test `conduit/plugins/importers/filereader/fileReadWrite_test.go`


conduit/plugins/exporters/filewriter/util.go
```diff
@@ -1,52 +1,96 @@
...
-	prettyHandle.ErrorIfNoField = json.CodecHandle.ErrorIfNoField
-	prettyHandle.ErrorIfNoArrayExpand = json.CodecHandle.ErrorIfNoArrayExpand
-	prettyHandle.Canonical = json.CodecHandle.Canonical
-	prettyHandle.RecursiveEmptyCheck = json.CodecHandle.RecursiveEmptyCheck
-	prettyHandle.Indent = json.CodecHandle.Indent
-	prettyHandle.HTMLCharsAsIs = json.CodecHandle.HTMLCharsAsIs
-	prettyHandle.MapKeyAsString = true
-	prettyHandle.Indent = 2
-
-	jsonStrictHandle = new(codec.JsonHandle)
```
tzaffi, line 29: removing `jsonStrictHandle`

conduit/plugins/exporters/filewriter/util.go
```diff
@@ -55,41 +99,51 @@ func EncodeJSONToFile(filename string, v interface{}, pretty bool) error {
...
 
-	var handle *codec.JsonHandle
-	if pretty {
-		handle = prettyHandle
-	} else {
-		handle = jsonStrictHandle
+	return Encode(format, writer, v)
+}
+
+// Encode encodes an object to a writer using a given an EncodingFormat.
```
tzaffi, line 105:
```suggestion
// Encode an object to a writer using a given an EncodingFormat.
```

conduit/plugins/importers/filereader/README.md
```diff
@@ -10,16 +10,7 @@ config:
     # The directory is created if it doesn't exist. If no directory is provided
     # blocks are written to the Conduit data directory.
     #block-dir: "/path/to/directory"
-    
-    # RetryDuration controls the delay between checks when the importer has
```
tzaffi, line 14: removing retries from the file plugins

conduit/plugins/importers/filereader/test_resources/conduit_data/conduit.yml
```diff
@@ -0,0 +1,72 @@
...
+        # formatting and should accept one number for the round.
+        filename-pattern: "%[1]d_block.msgp.gz"
+
+# Zero or more processors may be defined to manipulate what data
+# reaches the exporter.
+processors:
+
+# An exporter is defined to do something with the data.
+exporter:
+  name: "file_writer"
```
tzaffi, line 46:
```suggestion
  name: file_writer
```

conduit/plugins/importers/filereader/test_resources/filereader_blocks/0_block.msgp.gz
tzaffi, line 1: These test fixtures were generated using the block generator scenario [config.allmixed.small.yml](https://github.com/algorand/go-algorand/blob/master/tools/block-generator/scenarios/config.allmixed.small.yml)

conduit/plugins/exporters/filewriter/util.go
```diff
@@ -1,52 +1,96 @@
...
+	jsonPrettyHandle.HTMLCharsAsIs = json.CodecHandle.HTMLCharsAsIs
+	jsonPrettyHandle.MapKeyAsString = true
+	jsonPrettyHandle.Indent = 2
 }
 
-// EncodeJSONToFile is used to encode an object to a file. If the file ends in .gz it will be gzipped.
-func EncodeJSONToFile(filename string, v interface{}, pretty bool) error {
-	var writer io.Writer
+// ParseFilenamePattern parses a filename pattern into an EncodingFormat and gzip flag.
+func ParseFilenamePattern(pattern string) (EncodingFormat, bool, error) {
```
winder, line 44: Do these need to be public? There is some code shared between the filewriter and filereader, that could explain why the Encode function is public.

conduit/plugins/exporters/filewriter/file_exporter_test.go
```diff
@@ -21,29 +20,45 @@ import (
 	sdk "github.com/algorand/go-algorand-sdk/v2/types"
 )
 
+const (
+	defaultEncodingFormat = MessagepackFormat
+	defaultIsGzip         = true
+)
+
```
winder, line 27: These don't seem to be used. I think the defaults derived from `filewriter.FilePattern`.
```suggestion
```

conduit/plugins/exporters/filewriter/file_exporter_test.go
```diff
@@ -155,55 +170,67 @@ func sendData(t *testing.T, fileExp exporters.Exporter, config string, numRounds
...
-		assert.NotContains(t, string(blockBytes), " 0: ")
+	patterns := []string{
+		"%[1]d_block.json",
+		"%[1]d_block.json.gz",
+		"%[1]d_block.msgp",
+		"%[1]d_block.msgp.gz",
+	}
+	for _, pattern := range patterns {
+		pattern := pattern
+		t.Run(pattern, func(t *testing.T) {
```
winder, line 181: :+1:

conduit/plugins/importers/filereader/filereader_test.go
```diff
@@ -22,10 +21,13 @@ import (
 	sdk "github.com/algorand/go-algorand-sdk/v2/types"
 )
 
+const (
+	defaultEncodingFormat = filewriter.MessagepackFormat
+	defaultIsGzip         = true
+)
+
```
winder, line 28: Also unused?
```suggestion
```

conduit/plugins/importers/filereader/test_resources/conduit_data/conduit.yml
```diff
@@ -0,0 +1,72 @@
+# Log verbosity: PANIC, FATAL, ERROR, WARN, INFO, DEBUG, TRACE
+log-level: INFO
```
winder, line 2: I'd consider making this programmatically, or removing as much as possible so that there's less opportunity for things to become stale.

conduit/plugins/importers/filereader/filereader.go
```diff
@@ -66,14 +66,28 @@ func (r *fileReader) Init(ctx context.Context, _ data.InitProvider, cfg plugins.
...
+	}
 
 	return nil
 }
 
+// GetGenesis returns the genesis. Is is assumed that
+// the genesis file is available __in addition to__ the round 0 block file.
+// This is because the encoding assumed for the genesis is different
+// from the encoding assumed for blocks.
+// TODO: handle the case of a multipurpose file that contains both encodings.
```
winder, line 81: Could you elaborate on this TODO in the comment? As it is, I'm not sure exactly what you have in mind.

Also, the canonical genesis file is json. Maybe we should always write it that way?

conduit/plugins/metadata.go
```diff
@@ -13,11 +23,31 @@ type PluginType string
...
-	Processor = "processor"
+	Processor PluginType = "processor"
 
 	// Importer PluginType
-	Importer = "importer"
+	Importer PluginType = "importer"
 )
+
+// GetConfig creates an appropriate plugin config for the type.
+func (pt PluginType) GetConfig(cfg data.NameConfigPair, dataDir string) (PluginConfig, error) {
```
winder, line 36: What is this for? I'm having a hard time figuring out where it's used. There are already several patterns for creating test configurations in the other plugin tests, I think it would be better to use/refine one of those rather than adding a new scheme.

winder (changes requested):
The actual feature looks fine, nice improvement. I have some misc questions and comments about the refactoring and some potential dead code.

conduit/plugins/exporters/filewriter/file_exporter_test.go
tzaffi: These are used below in the (repurposed) unit test `TestPatternDefault` and in the admittedly silly'ish `TestDefaults` which is just a sanity check for the future in case the defaults change.

conduit/plugins/importers/filereader/filereader.go
tzaffi: What I mean in the **TODO** is that in principle, it's possible to have a block 0 which includes both the genesis information as well as the block information. But you hint that this may not be such a great idea and we should probably just assume we have a `genesis.json` since that's canonical. Does it actually make sense to have a round 0? Should we eliminate the concept?

conduit/plugins/metadata.go
tzaffi: I'm only using it in `fireReadWrite_test.go`, and also in the recently renamed `pipelineImpl.configWithLogger()`. I can easily recreate the logic in the test and revert this new function. I'll go ahead and do that, but I'm going to keep the added typing in this file to the constants `Exporter, Processor` and `Importer` as I believe that was the original intent.

conduit/plugins/importers/filereader/filereader.go
tzaffi: Plan:
* remove TODO
* file exporter/importer ASSUME genesis.json
* continue with block 0 in appropriate encoding

conduit/plugins/metadata.go
winder: Thanks, yes please revert this and add a test helper if necessary. I think this is unique to the filereader/filewriter e2e test, so it probably doesn't need to be a generally available utility.

conduit/plugins/exporters/filewriter/util.go
tzaffi: It's used by by the filereader as well. I was thinking of creating a common library. But I think this is fine as is for now.

conduit/plugins/importers/filereader/filereader.go
tzaffi: 497df82430e75ed109836893133b0bd5d82d1c1f

conduit/plugins/importers/filereader/test_resources/conduit_data/conduit.yml
tzaffi: d7fd94890b1b6fa5dc7bd8b54eb809f5f1abcca6

conduit/plugins/metadata.go
tzaffi: e08d93a555bf9621b4d7f9eab572b4fd1a4e3bcf and e9af0bd7fc1ff5cdb73b6ab81cf46f8733f1b882

conduit/plugins/importers/filereader/fileReadWrite_test.go
```diff
@@ -0,0 +1,240 @@
...
+		err = os.MkdirAll(config.DataDir, os.ModePerm)
+		require.NoError(t, err)
+	}
+
+	return config
+}
+
+// TestRoundTrip tests that blocks read by the filereader importer
+// under the msgp.gz encoding are written to identical files by the filewriter exporter.
+// This includes both a genesis block and a round-0 block with differend encodings.
```
tzaffi, line 142:
```suggestion
// This includes both a genesis block and a round-0 block with different encodings.
```

conduit/plugins/importers/filereader/filereader_test.go
```diff
@@ -57,7 +64,9 @@ func initializeTestData(t *testing.T, dir string, numRounds int) sdk.Genesis {
 		Timestamp:   1234,
 	}
 
-	err := filewriter.EncodeJSONToFile(path.Join(dir, "genesis.json"), genesisA, true)
+	genesisFilename := filewriter.GenesisFilename
```
shiqizng, line 67: nit: var assignment seems unnecessary here.

conduit/plugins/exporters/filewriter/util.go
```diff
@@ -55,41 +79,51 @@ func EncodeJSONToFile(filename string, v interface{}, pretty bool) error {
...
+// Encode an object to a writer using a given an EncodingFormat.
+func Encode(format EncodingFormat, writer io.Writer, v interface{}) error {
+	var handle codec.Handle
+	switch format {
+	case JSONFormat:
+		handle = jsonPrettyHandle
+	case MessagepackFormat:
+		handle = msgpack.LenientCodecHandle
+	default:
+		return fmt.Errorf("EncodeToFile(): unhandled format %d", format)
```
shiqizng, line 94:
```suggestion
		return fmt.Errorf("Encode(): unhandled format %d", format)
```

conduit/plugins/exporters/filewriter/util.go
```diff
@@ -1,52 +1,76 @@
...
+		blockFormat = JSONFormat
+	} else {
+		return UnrecognizedFormat, false, fmt.Errorf("unrecognized export format: %s", originalPattern)
+	}
+
+	return blockFormat, gzip, nil
+}
 
+// EncodeToFile enocods an object to a file using a given format and possible gzip compression.
+func EncodeToFile(filename string, v interface{}, format EncodingFormat, isGzip bool) error {
```
shiqizng, line 65: can the format be inferred from the filename extension?

conduit/plugins/importers/filereader/filereader_test.go
```diff
@@ -70,7 +79,7 @@ func initializeTestData(t *testing.T, dir string, numRounds int) sdk.Genesis {
 			Certificate: nil,
 		}
 		blockFile := path.Join(dir, fmt.Sprintf(filewriter.FilePattern, i))
-		err = filewriter.EncodeJSONToFile(blockFile, block, true)
+		err = filewriter.EncodeToFile(blockFile, block, defaultEncodingFormat, defaultIsGzip)
```
shiqizng, line 82: do we need to define `defaultEncodingFormat`, and `defaultIsGzip` are consts in this file? It looks like we could replace these consts with `filewrite.MessagepackFormat` and `true` and remove `TestDefaults`.

conduit/plugins/exporters/filewriter/util_test.go
```diff
@@ -1,13 +1,109 @@
...
+			blockFormat: UnrecognizedFormat,
+			err:         "unrecognized export format",
+		},
+		{
+			name:        "invalid - no extension",
+			format:      "%d_block",
+			gzip:        false,
+			blockFormat: UnrecognizedFormat,
+			err:         "unrecognized export format",
+		},
```
shiqizng, line 88: add another case for format without the %d specifier?

conduit/plugins/importers/filereader/filereader_test.go
tzaffi: b18d85b075f1132663cebaeaa4a520b9a8756087

conduit/plugins/exporters/filewriter/util.go
tzaffi: Here I'm proposing to separate the responsibility of deciding which encoding from which filename extension. The exporter's `Init()` parses the configured (or default) file pattern and sets its `format` field. So the caller of `EncodeToFile()` has to ensure that it's using the appropriate extension for a given format.

For this exporter's `Receive()`, the filename is constructed appropriately via:
```go
path.Join(exp.cfg.BlocksDir, fmt.Sprintf(exp.cfg.FilenamePattern, exportData.Round()))
```

conduit/plugins/importers/filereader/filereader_test.go
tzaffi: I'm aiming for an explicit declaration that these are the defaults, in case this changes in the future. However, `TestDefaults()` isn't actually asserting that the defined defaults are accurate. I've fixed this here: fb7ea14e340df39f64b1c7281a19fb36a7f1ad40

conduit/plugins/exporters/filewriter/util_test.go
tzaffi: I originally did, but unfortunately it doesn't fail, just gives a really gnarly file name that likely wouldn't get saved:

https://go.dev/play/p/S4QyqmUJVp6

I could add some explicit checks, but it's not obvious to me exactly what they should check for.

shiqizng: that's fine as long as we document the expected filename pattern.

conduit/plugins/importers/filereader/fileReadWrite_test.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	return config
+}
+
+// TestRoundTrip tests that blocks read by the filereader importer
+// under the msgp.gz encoding are written to identical files by the filewriter exporter.
+// This includes both a genesis block and a round-0 block with different encodings.
+func TestRoundTrip(t *testing.T) {
+	cleanArtifacts(t)
```
shiqizng, line 144: maybe use  t.TempDir()?

shiqizng (approved):

conduit/plugins/importers/filereader/fileReadWrite_test.go
tzaffi: I was considering a tempdir but since I wanted to reuse the artifacts for integration testing, I found this a bit easier -as you know exactly where the artifacts are landing before the test begins. But I'm open to further discussion.

winder: This seems reasonable

conduit/plugins/exporters/filewriter/README.md
```diff
@@ -2,7 +2,10 @@
 
 Write block data to files. This plugin works with the file rerader plugin to create a simple file-based pipeine.
 
+The genesis is always exported to a plain JSON file named `genesis.json` regardless of the `FilenamePattern`.
```
winder, line 5:
```suggestion
The genesis file is always exported to a plain JSON file named `genesis.json` regardless of the `FilenamePattern`.
```

conduit/plugins/exporters/filewriter/file_exporter.go
```diff
@@ -51,20 +57,34 @@ func (exp *fileExporter) Init(_ context.Context, initProvider data.InitProvider,
...
+	if err != nil && !errors.Is(err, os.ErrExist) {
+		// Ignore mkdir err if the dir exists (case errors.Is(err, os.ErrExist))
 		return fmt.Errorf("Init() error: %w", err)
 	}
+
 	exp.round = uint64(initProvider.NextDBRound())
-	return err
+
+	genesis := initProvider.GetGenesis()
+	genesisPath := path.Join(exp.cfg.BlocksDir, GenesisFilename)
```
winder, line 79: This was missing before?

winder (approved):

conduit/plugins/exporters/filewriter/file_exporter.go
tzaffi: Yeah

PR #144, opened 2023-08-23
Add credentials for telemetry

algochoi:
## Summary

Adds telemetry credentials for our Opensearch dev cluster.

## Test Plan

Push telemetry events on startup on local conduit instance.

### DIFF
--- a/pkg/cli/internal/initialize/conduit.yml.example
+++ b/pkg/cli/internal/initialize/conduit.yml.example
@@ -37,3 +37,28 @@ processors:
 # An exporter is defined to do something with the data.
 exporter:
 %s
+<<<<<<< Updated upstream
+=======
+
+# Enable telemetry for conduit
+telemetry:
+<<<<<<< Updated upstream
+  enabled: false
+
+  # By default the following fields will be configured to send data to Algorand.
+  # To store your own telemetry events, they can be overridden.
+  # uri: ""
+  # index: ""
+  # username: ""
+  # password: ""
+=======
+    enabled: false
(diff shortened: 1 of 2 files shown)


tzaffi (approved):

pkg/cli/internal/initialize/conduit.yml.example
```diff
@@ -37,3 +37,14 @@ processors:
...
+# Enable telemetry for conduit
+telemetry:
+    enabled: false
+
+    # By default the following fields will be configured to send data to Algorand.
+    # To store your own telemetry events, they can be overridden.
+    # uri: "https://conduit-telemetry.deveks.algodev.network"
+    # index: "conduit-telemetry-testing"
+    # username: "conduit-telemetry"
+    # password: "ratchet-tahiti-hunker-plebs-hamburg-wiretap-aplomb-mongoose-roomful"
```
winder, line 50: Could you move this next to `metrics` so that all the plugins are configured at the end?

winder (changes requested):
Could you move the defaults into the code? That way they can be updated with a software update rather than going stale if we want to change them in the future.

algochoi:
> Could you move the defaults into the code? That way they can be updated with a software update rather than going stale if we want to change them in the future.

That makes sense since we will likely change the URI some time in the future. But moving the defaults in the code could make it harder for users to set their custom telemetry endpoints (changing source code and recompiling the binary seems a bit more cumbersome). Is this an okay tradeoff here?

Issue #141, opened 2023-08-18
Pipelining Followups


tzaffi:
## Problem

Some issues remaining at the end of #128. However, the long validation process made it safer to push fixes as followups

## Remove the `*pImpl` parameter from `Retries()` in `conduit/pipeline/common.go`
Per @winder 
> * breaking out the context.Context out into a parameter, I was specifically looking for it to make sure the wait is properly terminated.
> * also breaking out RetryCount and logger. It's already a pure function in how it's used, but the signature doesn't guarantee that it's pure.

## Urgency

Medium

PR #147, opened 2023-08-29
Pipelining followups

tzaffi:
# Summary

Addressing the [Pipelining Followups Issue](https://github.com/algorand/conduit/issues/141) along with the "CLI health endpoint bug" which I discovered while working on the block generator.

## The CLI Health Endpoint Bug
This bug was actually introduced by the [Pipelining PR](https://github.com/algorand/conduit/pull/128) which preceded the health endpoint, and in particular [this line](https://github.com/tzaffi/conduit/blob/9918073c60effa45f05c9266e53f500a5d672410/conduit/pipeline/pipeline.go#L684) which caused `Start()` itself to wait forever. As a consequence, the setup of the health endpoint was never reached while conduit was running.

## Issue

#141 

## TODO's

- [x] remove `retriesNoOutput()` and its test, unless a use case for _this PR_ comes up.

## Testing

Introducing `TestHealthEndpoint()` in `cli_test.go` for checking that the health endpoint is ON/OFF as configured. I verified that when the bug is re-introduced, the `API_ON` case fails as expected. For the purposes of the test, I also introduced a `noop importer` which waits 100 ms before returning in its `GetBlock()` method.


conduit/pipeline/common.go
```diff
@@ -74,9 +76,19 @@ func Retries[X pluginInput, Y pluginOutput](f func(x X) (Y, error), x X, p *pipe
 	return
 }
 
-// RetriesNoOutput applies the same logic as Retries, but for functions that return no output.
-func RetriesNoOutput[X pluginInput](f func(x X) error, a X, p *pipelineImpl, msg string) (time.Duration, error) {
-	_, d, err := Retries(func(x X) (empty, error) {
+// TODO: probly the following function and its unit test should be axed
+// retriesNoInput applies the same logic as Retries, but for functions that take no input.
```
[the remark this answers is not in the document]
tzaffi: @algochoi - I had originally added this because I wanted to retry the call to `GetGenesis()` (which takes no params) inside of the pipeline's `Init()`. But then I realized that retrying at that point is too late and you'd really have have the importer do that (and it can't access this package).

UPSHOT: I commit to deleting this function and its test before merging, unless there's a use case that crops up mighty soon.

conduit/pipeline/pipeline.go
```diff
@@ -696,8 +695,6 @@ func (p *pipelineImpl) Start() {
 			}
 		}
 	}(p.pipelineMetadata.NextRound)
-
-	<-p.ctx.Done()
```
tzaffi, line 700: This is the health endpoint bugfix.

conduit/plugins/importers/noop/noop_importer.go
```diff
@@ -0,0 +1,81 @@
...
+func (imp *noopImporter) Close() error {
+	return nil
+}
+
+func (imp *noopImporter) GetGenesis() (*sdk.Genesis, error) {
+	return &sdk.Genesis{}, nil
+}
+
+func (imp *noopImporter) GetBlock(rnd uint64) (data.BlockData, error) {
+	time.Sleep(sleepForGetBlock)
```
winder, line 64: Why the sleep?

winder (approved):
Approved pending your offer to remove the unused function.

conduit/plugins/importers/noop/noop_importer.go
tzaffi: Just to slow down the pipeline. For example, if you have a no-op for your importer and a file-writer for your exporter this limits the output to 10 blocks per second which is a little easier to manage than without having any sort of brakes on. Another option is to make this configurable, but that didn't strike me as ideal either since I'd like the no-op importer to have no config. LMK.

tzaffi:
> Approved pending your offer to remove the unused function.

d6358415c5c2cd598662697ab829b44977e7c3b3

algochoi (approved):

winder (approved):

PR #150, opened 2023-09-12
ci: Add gpg signature to checksum file.

winder:
## Summary

Sign checksum file as part of publishing a release.

The approach here follows a suggestion in the goreleaser docs where the signature is limited to the checksum file.

References:
* goreleaser docs: https://goreleaser.com/customization/sign/
* ghaction-import-gpg docs: https://github.com/crazy-max/ghaction-import-gpg

Repository configuration requires adding the GPG secret and passphrase.

## Test Plan

Tested the configuration in a private repo:
https://github.com/algorand/goreleaser-test/releases/tag/untagged-e548c44def7b11988fec

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -95,6 +95,9 @@ changelog:
       - '^docs:'
       - '^test:'
 
+signs:
+  - artifacts: checksum
+
 release:
   draft: true
   header: |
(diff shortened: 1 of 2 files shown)


.goreleaser.yaml
```diff
@@ -95,6 +95,9 @@ changelog:
       - '^docs:'
       - '^test:'
 
+signs:
+  - artifacts: checksum
```
onetechnical, line 99:
```suggestion
  - artifacts: all
```

Would this do detached signatures?

winder: The docs recommended using `checksum`, so thats what I went with here. It gives you transitive protection since the checksums would be signed and able to verify all other artifacts.

onetechnical (approved):
This looks fine for now, though we may want to modify this to do detached sigs as well.

PR #154, opened 2023-10-10
Set conduit data dir through env var

cshintov:
Use `CONDUIT_DATA_DIR` in the entrypoint script for docker.


winder (approved):
Looks great, thanks for the fix. If you'd like to contribute another PR, the Indexer image is virtually identical to this one. The same fix applies there.

cshintov:
Yup. Saw it. Was thinking to raise a PR for that as well. Thanks for the merge 🙂 !

cshintov:
Done. https://github.com/algorand/indexer/pull/1583

PR #162, opened 2024-01-06
Support to add custom consensus in conduit

scholtz:
This is dependant on https://github.com/algorand/go-algorand-sdk/pull/625

## Summary

Conduit cannot process any blocks for algorand blockchains with custom consensus. This 5 line PR fixes it.

## Test Plan

I tested it manually on aramidmain blockchain.


scholtz:
for the record in the current official conduit is still the bug that the custom consensus cannot be processed

image

PR #167, opened 2024-04-18
Allow catchpoints from the exact target round

jasonpaulos:
## Summary

https://github.com/algorand/go-algorand/issues/5352 has been fixed by https://github.com/algorand/go-algorand/pull/5798, and this PR undoes the workaround needed because of the original issue.

The original issue was that, if algod does fast catchup to round `X`, algod's REST API could not serve block `X`, since fast catchup did not also include block certificates. To work around this, conduit had to limit catchpoints to rounds `< X`, but now with this fixed in algod, the catchpoint round can be `<= X`.

## Test Plan

New unit test cases added


jasonpaulos:
Manually tested this change with a follower node and fast catchup, and I can confirm it worked as expected

gmalouf (approved):

PR #179, opened 2024-12-20
Incentives: Bump to v40 consensus supporting development versions of Go SDK and Indexer.

gmalouf:
## Summary

Support v40 consensus version and new heartbeat transaction type.

## Test Plan

Existing tests should pass.

### DIFF
--- a/conduit/plugins/processors/filterprocessor/gen/generate.go
+++ b/conduit/plugins/processors/filterprocessor/gen/generate.go
@@ -60,6 +60,8 @@ var ignoreTags = map[string]bool{
 	"txn.apsu": true,
 	"dt.lg":    true,
 	"dt.sa":    true,
+	// Filter out heartbeat fields that are not obvious how to filter upon
+	"txn.hb": true,
 }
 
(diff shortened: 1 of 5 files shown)


conduit/plugins/processors/filterprocessor/gen/generate.go
```diff
@@ -60,6 +60,8 @@ var ignoreTags = map[string]bool{
 	"txn.apsu": true,
 	"dt.lg":    true,
 	"dt.sa":    true,
+	// Filter out heartbeat fields that are not obvious how to filter upon
+	"txn.hb": true,
```
jannotti, line 64: I have no idea what's going on here, so I'm guessing here. But is seems like the hb address is a worthwhile and understandable field. I guess the rest of the hb fields are not very interesting.

algorandskiy (approved):
I guess generally there is no interest in heartbeat txns? so LGTM

gmalouf:
> I guess generally there is no interest in heartbeat txns? so LGTM

I am happy to add filter processors for heartbeat address if people end up wanting them, but can be follow-up.

